PR #12, opened 2015-07-13
CP-766 More robust module lifecycle with examples

trentgrover-wf:
## Issue
- Module lifecycle warranted a bit of expansion:
  - A module must be aware of its children so that it can manage the lifecycle of its underlings
  - The consumer of a module must be able to track its load / unload state
  - A module deep in an app's hierarchy must be able to prevent unloading of that module (e.g. 'Are you sure you want to close this panel with important unsaved changes?')
- We don't have any really solid examples of why this module pattern should be employed
## Changes

**Source:**
- Expanded lifecycle module:
  - Added an internally managed list of child modules
  - Added broadcast streams for module load / unload events
  - Added explicit loadModule method to register a new child module with 'this' parent module
  - Expanded shouldUnload functionality so that if works hierarchically with a network of child modules
  - shouldUnload returns a bool and a list of strings (intended to contain explanatory messages from modules that have rejected unload)
- Added examples for many module use cases:
  - Screencast: http://screencast.com/t/zUB9gkCrN
  - Parent panel - dynamically loads / unloads child modules as needed in response to nav bar selection
  - Basic module - minimal renderable module boilerplate
  - Flux module - minimal implementation of a flux based module (action + store + component)
  - Reject module - checkbox controls the module’s shouldUnload result, demonstrating a module’s ability to prevent unloading no matter how deep it lies in the hierarchy
  - Data Load (async) - demonstrates a recommended pattern for a module that needs to load data.  Module async requests data immediately onLoad, but provides a UI component that is immediately renderable without data (shows a loading indicator and updates as async data resolves)
  - Data Load (blocking) - similar to async data load example, but this example blocks rendering of the module until initial data is ready.  I wouldn’t recommend this pattern, but it requires much less code (no flux)
  - Deferred - module gets its data from a deferred dependency.  demonstrates the interface - implementation split required to get this to work with Dart’s static typing system.  deferred dependency loading is a good use of the module onLoad method
  - Lifecycle Echo - module that just echos all available module lifecycle events to the dev console.  demonstrates how to access / use all of the available hooks
  - All of them - a module that loads all of the other modules as children.  demonstrates the hierarchical capabilities of the module pattern.  includes button to explicitly unload specific child modules
  - Recursive - loads the whole panel example module as a child - recursively.  demonstrates the hierarchical capabilities of the module lifecycle patterns / parent-child relationships.  (e.g. load multiple levels of this example and use a deep version of the Reject module to prevent reload of the page or selection of any other module)

**Tests:**
- added unit tests for lifecycle_module
## Areas of Regression
- mostly everywhere that uses w_module.  shouldn't really cause any regressions because most, if not all, of our modules are not actually leveraging the lifecycle
## Testing
- Run the example
- CI passes
## Code Review

@maxwellpeterson-wf
@evanweible-wf
@dustinlessard-wf 
@jayudey-wf

### DIFF
--- a/example/lib/src/panel/deferred_heavyLifter_implementation.dart
+++ b/example/lib/src/panel/deferred_heavyLifter_implementation.dart
@@ -40,7 +40,7 @@ class RealLifter implements HeavyLifter {
         'Bubble Buddy',
         'DoodleBob'
       ];
-    } else if (_division == HeavyLifterDivision.FEATHERWEIGHT) {
+    } else if (_division == HeavyLifterDivision.WELTERWEIGHT) {
       return [
         'Philip J. Fry',
         'Turanga Leela',


travisreed-wf:
running rosie v2 on this pull

PR #14, opened 2015-07-21
CP-980 Add a DeferredModule generator

evanweible-wf:
## Issue
It should be possible to consume any module as a deferred module. And it should be easy.

##### Module Consumer
The only difference should be which version of the module a consumer imports - either the regular one, or the deferred one. The typing and interaction should be identical.

##### Module Creator
Creating a deferred version of a module comes down to the following steps:
- importing the actual implementations as deferred libraries
- loading the above deferred libraries during the `load` lifecycle event
- preventing access to API/components/events until they have been loaded

This is a lot of boilerplate that will generally be the same across every module. There should be a way to generate it!

## Changes
**Docs:**
- Adds a walkthrough for generating a deferred module: https://github.com/Workiva/w_module/blob/deferred-module-gen/docs/generating_a_deferred_module.md

**Source:**
- Adds a `DeferredModuleGenerator` that uses the `source_gen` library to generate a deferred module class for a given module. It also generates abstract views of the API, components, and events classes if available.
- Adds a `DeferredModule` annotation that should be used to inform the generator about the locations of the relevant module classes.

**Tests:**
- None, yet.

## Areas of Regression
- n/a

## Testing
Walk through the deferred module generation guide with any module. It should generate a valid deferred module that can be used in place of the regular module with zero changes to consumption.

## TODO
- [x] Don't require API/components/events classes.
- [x] Support parameters in the module constructor
  - The API/components/events classes don't have this issue because we don't need to construct them, the actual module will take care of that
  - The deferred module does need to be constructed, and it should work just like the regular module.
- [x] Support named constructors. Same reason as above.
- [ ] Support static class members.
- [ ] Rewrite the example to be understandable/realistic.
- [ ] Add source gen utils to our `dart_dev` package to make the generation step easier. w_ui_platform has a bunch of relevant code that should be used.

## Caveats
~~As far as I can tell, the analyzer package (used by source_gen) only gives you the most basic type information. For example, if an API method had a return type of `Future<bool>`, the typing within the generic is lost, leaving only `Future`. This is particularly an issue for the API and Events classes because they will be dealing with typed Futures and Streams, but that specificity will be lost when generating the deferred version.~~ Generics work!

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
fyi: @jayudey-wf @shanesizer-wf


evanweible-wf:
Some additional todos/points of concern:

- [ ] Fine-grained control over which parts of the Module are deferred and which parts should be available up-front. Use case: deferred modules will likely want to provide a simple component (like a sidebar icon) that can trigger the loading of itself. This might still need to be managed at a higher level to avoid the module performing logic before being officially loaded. Perhaps a component like this could be exposed as a static property on the module and the parent can render the component and pass in a callback that calls `.loadChildModule()`.

evanweible-wf:
Went with a transformer approach instead: #40

PR #16, opened 2015-07-22
CP-785 Add DispatchKey to control event dispatching

evanweible-wf:
## Issue

Events are currently hard to control because they need to be read-only when exposed outside of a module.
## Changes

**Source:**
- Got rid of `Event.fromStream()` constructor **(breaking change)**
- Add a `DispatchKey` and associate events with a certain key upon construction
- Make `Event` class invokable to enable easy event dispatching
  - The dispatch key will be required to successfully dispatch an event

**Test:**
- Relevant tests added
## Areas of Regression
- Event dispatching
## Testing
- Tests pass

### DIFF
--- a/lib/src/event.dart
+++ b/lib/src/event.dart
@@ -3,12 +3,45 @@ library w_module.src.event;
 import 'dart:async';
 
 class Event<T> extends Stream<T> {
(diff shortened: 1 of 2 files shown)


travisreed-wf:
running rosie v2 on this pull

PR #39, opened 2015-12-02
dart 1.13

jayudey-wf:


lib/src/lifecycle_module.dart
```diff
@@ -67,9 +67,9 @@ abstract class LifecycleModule {
   /// Calls the onLoad() method, which can be implemented on a Module.
   /// Executes the willLoad and didLoad event streams.
   Future load() async {
-    _willLoadController.add(this);
+    await _willLoadController.add(this);
     await onLoad();
-    _didLoadController.add(this);
+    await _didLoadController.add(this);
```
timmccall-wf: The `await onLoad()` in the previous code is necessary to ensure that the events are properly timed. Not sure why you needed to await the addition of the payload to the event streams. What was breaking without this change?

lib/src/lifecycle_module.dart
```diff
@@ -77,12 +77,12 @@ abstract class LifecycleModule {
   Future loadChildModule(LifecycleModule newModule) async {
     newModule.didLoad.listen((_) {
       _childModules.add(newModule);
-      _didLoadChildModuleController.add(newModule);
     });
     newModule.willUnload.listen((_) {
       _childModules.remove(newModule);
     });
     await newModule.load();
+    await _didLoadChildModuleController.add(newModule);
```
timmccall-wf: I think the previous code was more correct/better composed, with the event dispatch inside the `newModule.didLoad` handler. But again, I am very curious why Dart 1.13 necessitated this change.

lib/src/lifecycle_module.dart
jayudey-wf: it seemed in the prior release of dart that the `load()` future wouldn't resolve itself until all of the events had been propagated to the streams and listeners, where as after 1.13.0 the future would resolve itself prior the listener actually receiving the data

lib/src/lifecycle_module.dart
jayudey-wf: i moved this here because I couldn't successfully await this during the listener above

trentgrover-wf: for readability sake, just completely kill the entire `newModule.didLoad.listen` block so the bottom code chunk can be
```
await newModule.load();
_childModules.add(newModule);
await _didLoadChildModuleController.add(newModule);
```

lib/src/lifecycle_module.dart
trentgrover-wf: don't know why it should work this way, but it definitely does...an alternative that also works is to replace the async / await syntax with completers:
https://github.com/Workiva/w_module/pull/41
any votes on which approach to use?

evanweible-wf: My guess is that it works because the `await` syntax causes that line to be wrapped in another async block when compiled, which effectively waits an extra tick. But `StreamSink.add()` definitely returns void (https://api.dartlang.org/1.13.0/dart-async/EventSink/add.html) so adding an `await` isn't correct, it just produces a side effect that happens to work. For that reason, I'd vote for #41.

jayudey-wf: agreed, i'll shut down this pr

PR #48, opened 2016-07-21
REF-1068 Add additional lifecycle methods

georgelesica-wf:
[Jira Ticket](https://jira.atl.workiva.net/browse/REF-1068)

## Problem

Lux uses a pattern whereby they add their components in an `onDidLoadChildModule` method that gets called automagically when a child module is loaded. This seems to work out very nicely and results in clean, easy-to-read code.

However, since we don't support this directly in the SDK or in w_module it is left up to consumers to implement the boilerplate and many consumers might not pick up on it unless they read the Lux source code. Additionally, and this is fairly minor, the only way to wire this up at the level of an Experience is to listen to the appropriate stream, which is slightly complex due to the need to cancel stream subscriptions.

I originally set out to add this method to the `Experience` in the wdesk_sdk repo, but it was suggested that this method, and similar child lifecycle methods might be useful on the `LifecycleModule` itself.

## Solution

Add support for this pattern.

## Potential Regressions

None?

## Tests

Added and/or updated.

## QA / +10

1. Checkout the branch.
2. Update dependencies.
3. Unit tests should pass.
4. Pull this branch into the `wdesk_examples` repo and make sure everything still works.
5. Pull this branch into `docs`, get rid of their homebrew stream subscription that makes this pattern work, and verify that their app still works (specifically, you should be able to create a document). Here's a branch with these changes already made: https://github.com/Workiva/docs/tree/test-ref-1068.

## FYI / FYA / FYR

@trentgrover-wf @maxwellpeterson-wf @evanweible-wf @jayudey-wf @Workiva/ref-pp

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -44,10 +44,30 @@ abstract class LifecycleModule {
   StreamController<LifecycleModule> _didUnloadController;
   Stream<LifecycleModule> get didUnload => _didUnloadController.stream;
 
+  /// Event dispatched at the beginning of module.loadChildModule() logic
+  StreamController<LifecycleModule> _willLoadChildModuleController;
+  Stream<LifecycleModule> get willLoadChildModule =>
+      _willLoadChildModuleController.stream;
+
   /// Event dispatched at end of module.loadChildModule() logic
   StreamController<LifecycleModule> _didLoadChildModuleController;
   Stream<LifecycleModule> get didLoadChildModule =>
       _didLoadChildModuleController.stream;
+
+  /// Event dispatched before a child module is unloaded
+  StreamController<LifecycleModule> _willUnloadChildModuleController;
+  Stream<LifecycleModule> get willUnloadChildModule =>
+      _willUnloadChildModuleController.stream;
+
+  /// Event dispatched after a child module is unloaded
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Raven
Number of Findings: 0

lib/src/lifecycle_module.dart
```diff
@@ -80,12 +106,30 @@ abstract class LifecycleModule {
   /// Public method to async load a child module and register it
   /// for lifecycle management.
   Future loadChildModule(LifecycleModule newModule) {
+    if (_childModules.contains(newModule)) {
```
georgelesica-wf: This keeps us from listening to the child's event streams more than once and ensures that the subscriptions get fully canceled when the child module unloads.

todbachman-wf:
+1

bencampbell-wf:
+1

bencampbell-wf:
@jayudey-wf ready for merge.

evanweible-wf:
+1 Looks good

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - performed and documented by Tod
- [ ] All unit tests pass 

Merging into master.

PR #55, opened 2016-09-22
REF-1267 Add suspend / resume semantics

georgelesica-wf:
## Problem

We need some concept of suspend / resume for other projects. It seems to make sense to add it here, at the lowest level.

## Solution

Add it. I added a few new bits:

  * `isLoaded` - whether the module is loaded, if they're allowed to be toggled, this seems like a good idea
  * `isSuspended` - this can be toggled, so it's nice to have a getter
  * `onSuspend` - called when we suspend
  * `onResume` - called when we resume
  * `willSuspend`, `didSuspend` - dispatched before and after we suspend
  * `willResume`, `didResume` - dispatched before and after we resume

I also added a `Logger` and added warnings in the following cases:

  * Loading an already-loaded module
  * Unloading a modules that has already been unloaded or was never loaded to begin with
  * Suspending an already-suspended module
  * Suspending a module that is not loaded
  * Resuming a module that is not suspended
  * Resuming a module that is not loaded

I updated the README to reflect the new features.

## Potential Regressions

Module load / unload.

## Tests

Added and/or updated.

## QA / +10

1. Checkout the branch.
2. Update dependencies.
3. Unit tests should pass.
4. CI passes.
5. No errors in examples.

## FYI

@Workiva/ref-pp @evanweible-wf @maxwellpeterson-wf @trentgrover-wf @dustinlessard-wf

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,23 +284,29 @@ Many examples of `Module` lifecycle behavior and manipulation can be found in th
 
 ### Lifecycle Methods
 
-`Module` exposes just three lifecycle methods that external consumers should use to trigger loading and unloading
+`Module` exposes five lifecycle methods that external consumers should use to trigger loading and unloading
 behavior:
 
 Method         | Description
 -------------- | ---------------------------------
 `load`         | Triggers loading of a `Module`.  Internally, this executes the module's `onLoad` method and dispatches the `willLoad` and `didLoad` events.  Returns a future that completes once the module has finished loading.
-`shouldUnload` | Returns the unloadable state of the `Module` as a `ShouldUnloadResult`.  Internally, this executes the module's `onShouldUnload` method.
-`unload`       | Triggers unloading of a `Module`.  Internally, this executes the module's `shouldUnload` method, and, if that completes successfully, executes the module's `onUnload` method. If unloading is rejected, this method will complete with an error.
+`suspend`      | Suspends the module and all child modules. While suspended modules should make themselves lightweight and avoid making network requests.
+`resume`       | Brings the module and all child modules out of suspension. Upon resuming, modules should go back to business as usual.
+`shouldUnload` | Returns the unloadable state of the `Module` and its child modules as a `ShouldUnloadResult`.  Internally, this executes the module's `onShouldUnload` method.
+`unload`       | Triggers unloading of a `Module` and all of its child modules.  Internally, this executes the module's `shouldUnload` method, and, if that completes successfully, executes the module's `onUnload` method. If unloading is rejected, this method will complete with an error.
 
 ### Lifecycle Events
 
 `Module` also exposes lifecycle event streams that an external consumer can listen to:
 
-Method         | Description
+Event          | Description
 -------------- | ---------------------------------
 `willLoad`     | Dispatched at the beginning of the module's `load` logic.
 `didLoad`      | Dispatched at the end of the module's `load` logic.
+`willSuspend`  | Dispatched at the beginning of the module's `suspend` logic.
+`didSuspend`   | Dispatched at the end of the module's `suspend` logic.
+`willResume`   | Dispatched at the beginning of the module's `resume` logic.
+`didResume`    | Dispatched at the end of the module's `resume` logic.
 `willUnload`   | Dispatched at the beginning of the module's `unload` logic.
 `didUnload`    | Dispatched at the end of the module's `unload` logic.
 
@@ -311,6 +317,8 @@ Internally, `Module` contains methods that can be overridden to customize module
 Method           | Description
 ---------------- | ---------------------------------
 `onLoad`         | Executing during the module's `load` logic.  Custom logic for initializing child modules, service access, event listeners, etc. should be implemented here.  Deferred module loading behavior can also be hidden from consumers via this method.
+`onSuspend`      | Executed during the module's `suspend` logic. This method should be used to modify module behavior while suspended.
+`onResume`       | Executed during the module's `resume` logic. This method should be used to put the module back into its normal mode of operation.
 `onShouldUnload` | Executed during the module's `shouldUnload` logic.  Custom logic that blocks module unloading under certain conditions should be implemented here.
 `onUnload`       | Executed during the module's `unload` logic.  Custom module clean up logic should be implemented here.  Unfortunately, the nature of web browsers is such that module `unload` logic is not guaranteed to be executed under all conditions (browser or tab close), so mission critical logic should not reside here.
 
@@ -318,10 +326,19 @@ Method           | Description
 
 `Module` also supports hierarchical application of the standard lifecycle through child modules:
 
-Method               | Description
+Method                | Description
+--------------------  | ---------------------------------
+`loadChildModule`     | Loads a child module and registers it with the current module for lifecycle management.
+`willLoadChildModule` | Dispatched at the beginning of the child module's `load` logic.
+`didLoadChildModule`  | Dispatched at the end of the child module's `load` logic.
+
+### Getters
+
+Getter               | Description
 -------------------- | ---------------------------------
-`loadChildModule`    | Loads a child module and registers it with the current module for lifecycle management.
-`didLoadChildModule` | Dispatched at the end of the child module's `load` logic.
+`childModules`       | An iterable of child modules.
+`isLoaded`           | A boolean that indicates whether the module is current loaded.
+`isSuspended`        | A boolean that indicates whether the module is currently suspended. This will always be false when the module is not loaded.
 
 
 ---
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -16,6 +16,8 @@ library w_module.src.lifecycle_module;
 
 import 'dart:async';
 
+import 'package:logging/logging.dart';
+
 /// Intended to be extended by most base module classes in order to provide a
 /// unified lifecycle API.
 abstract class LifecycleModule {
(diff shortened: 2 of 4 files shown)


aviary-wf:
## Raven
Number of Findings: 0

lib/src/lifecycle_module.dart
```diff
@@ -83,23 +107,45 @@ abstract class LifecycleModule {
         new StreamController<LifecycleModule>.broadcast();
     _didUnloadChildModuleController =
         new StreamController<LifecycleModule>.broadcast();
+    _willSuspendController = new StreamController<LifecycleModule>.broadcast();
+    _didSuspendController = new StreamController<LifecycleModule>.broadcast();
+    _willResumeController = new StreamController<LifecycleModule>.broadcast();
+    _didResumeController = new StreamController<LifecycleModule>.broadcast();
+
+    _logger = new Logger(name);
```
maxwellpeterson-wf: Maybe this should be `_logger = new Logger('w_module $name');` ?

maxwellpeterson-wf: Or just `_logger = new Logger('w_module');` and we include the name in each message:

```dart
_logger.warning('Module "$name" is already loaded, cannot load.');
```

lib/src/lifecycle_module.dart
```diff
@@ -83,23 +107,45 @@ abstract class LifecycleModule {
...
-      _didLoadController.add(this);
+    if (!_isLoaded) {
+      _willLoadController.add(this);
+      onLoad().then((_) {
+        _didLoadController.add(this);
+        _isLoaded = true;
+        completer.complete();
+      });
+    } else {
+      _logger.warning('Module is already loaded, cannot load.');
```
evanweible-wf: Could you include a `w_module: ` prefix as well as the module name? Should be available via `this.name`

lib/src/lifecycle_module.dart
```diff
@@ -159,25 +206,79 @@ abstract class LifecycleModule {
...
+            _isSuspended = false;
+            completer.complete();
+          });
+        });
+      } else {
+        // reject with shouldUnload messages
+        throw new ModuleUnloadCanceledException(canUnload.messagesAsString());
+      }
+    } else {
+      _logger.warning('Module is already unloaded, cannot unload.');
```
evanweible-wf: Same request here

lib/src/lifecycle_module.dart
evanweible-wf: Oh, didn't even notice that you were constructing the module with the `name` property. But like Max commented above it would be nice to have `w_module` in there as well

pubspec.yaml
```diff
@@ -16,6 +16,7 @@ dev_dependencies:
   dart_dev: "^1.0.0"
   dart_style: "^0.2.1"
   dartdoc: "^0.8.0"
+  logging: "^0.11.0"
```
evanweible-wf: This should be an actual dependency

georgelesica-wf: Hehe, oops!

georgelesica-wf: Oh wow, there were no real dependencies previously, I actually feel slightly bad for adding the first one!

georgelesica-wf: Oh wow, there were no real dependencies previously, I actually feel slightly bad for adding the first one!

evanweible-wf: Haha don't worry, logging is pretty universal. And I'm about to add another dependency anyway (meta)

georgelesica-wf:
@evanweible-wf @maxwellpeterson-wf addressed comments and squashed.

evanweible-wf:
+1, thanks @georgelesica-wf!

lib/src/lifecycle_module.dart
```diff
@@ -83,23 +107,45 @@ abstract class LifecycleModule {
...
   //--------------------------------------------------------
   // Public methods that can be used directly to trigger
   // module lifecycle / check current lifecycle state
   //--------------------------------------------------------
 
   /// Public method to trigger the loading of a Module.
+  ///
   /// Calls the onLoad() method, which can be implemented on a Module.
   /// Executes the willLoad and didLoad event streams.
+  /// TODO: Do we want to support re-loading of child modules here?
```
bencampbell-wf: Should this TODO be addressed?

georgelesica-wf: To me, it seems like a larger philosophical issue that will require some discussion across teams. It also seems like we can defer that discussion since it's gone this long without a resolution. I just wanted to make a note so we eventually deal with it.

If people would like to resolve it in this PR I'm game, we should probably talk on HC or GHO though.

@maxwellpeterson-wf @evanweible-wf thoughts?

README.md
```diff
@@ -323,6 +331,14 @@ Method               | Description
 `loadChildModule`    | Loads a child module and registers it with the current module for lifecycle management.
 `didLoadChildModule` | Dispatched at the end of the child module's `load` logic.
```
todbachman-wf: There is also a `willLoadChildModule` event. We might want to list that here as well.

lib/src/lifecycle_module.dart
todbachman-wf: I would vote for supporting it.

lib/src/lifecycle_module.dart
```diff
@@ -159,25 +206,80 @@ abstract class LifecycleModule {
...
+    }
+    return completer.future;
+  }
+
+  /// Public method to suspend the module.
+  ///
+  /// Suspend indicates to the module that it should go into a low-activity
+  /// state. For example, by disconnecting from backend services and unloading
+  /// heavy data structures.
+  Future suspend() {
```
todbachman-wf: #nit: maybe sort the methods?

README.md
```diff
@@ -284,25 +284,31 @@ Many examples of `Module` lifecycle behavior and manipulation can be found in th
...
 
-`Module` exposes just three lifecycle methods that external consumers should use to trigger loading and unloading
+`Module` exposes five lifecycle methods that external consumers should use to trigger loading and unloading
 behavior:
 
 Method         | Description
 -------------- | ---------------------------------
 `load`         | Triggers loading of a `Module`.  Internally, this executes the module's `onLoad` method and dispatches the `willLoad` and `didLoad` events.  Returns a future that completes once the module has finished loading.
 `shouldUnload` | Returns the unloadable state of the `Module` as a `ShouldUnloadResult`.  Internally, this executes the module's `onShouldUnload` method.
 `unload`       | Triggers unloading of a `Module`.  Internally, this executes the module's `shouldUnload` method, and, if that completes successfully, executes the module's `onUnload` method. If unloading is rejected, this method will complete with an error.
```
todbachman-wf: `unload` will unload all child modules too. `shouldUnload` will check the state of all child modules as well. It might be worth updating the descriptions of these two methods to note that.

lib/src/lifecycle_module.dart
georgelesica-wf: CP seems to put them in some kind of "logical" order so I tried to stick with that. Suspend comes before resume in that case because you always suspend before you can resume. When in Rome... I'm happy to change the order to whatever though, as long as it stays generally consistent throughout the library.

lib/src/lifecycle_module.dart
```diff
@@ -159,25 +206,80 @@ abstract class LifecycleModule {
...
-        _childModules.clear();
-        onUnload().then((_) {
-          _didUnloadController.add(this);
+    if (_isLoaded) {
+      ShouldUnloadResult canUnload = shouldUnload();
+      if (canUnload.shouldUnload) {
+        _willUnloadController.add(this);
+        Iterable<Future> unloadChildren = _childModules.map((c) => c.unload());
+        Future.wait(unloadChildren).then((_) {
+          _childModules.clear();
```
georgelesica-wf: I didn't change this code, just indented it because I added a guard.

lib/src/lifecycle_module.dart
evanweible-wf: The class is big enough now that I'd be okay with sorting it alphabetically. No strong preference either way.

georgelesica-wf: I reordered everything to be in a consistent "lifecycle"-y order, grouped into streams, methods, and callbacks. Seemed like a nice compromise, and keeps it easy to navigate.

lib/src/lifecycle_module.dart
georgelesica-wf: Ah, gotcha, re: the bit about the TODO above.

lib/src/lifecycle_module.dart
georgelesica-wf: Per @todbachman-wf let's hold off on this until we can discuss it. If we made the changes now we wouldn't leave consumers a good way to dispose a module, and the current state of this PR meets our current needs.

georgelesica-wf:
Squashed more updates. Ready for p1s.

todbachman-wf:
+1

evanweible-wf:
+1

dustinlessard-wf:
+1 lgtm

georgelesica-wf:
@jayudey-wf I think we're ready for merge here.

lib/src/lifecycle_module.dart
```diff
@@ -136,7 +182,55 @@ abstract class LifecycleModule {
...
+      Future.wait(_childModules.map((c) => c.suspend())).then((_) {
+        onSuspend().then((_) {
+          _didSuspendController.add(this);
+          _isSuspended = true;
+          completer.complete();
+        });
+      });
+    } else {
+      _logger.warning(
+          'Module "$name" is already suspended or not loaded, cannot suspend.');
```
trentgrover-wf: we have a way to check these conditions so why not use a message that's more specific?
```
_logger.warning('Module "$name" is ${_isLoaded ? 'already suspended' : 'not loaded'}, cannot suspend.');
```

lib/src/lifecycle_module.dart
```diff
@@ -136,7 +182,55 @@ abstract class LifecycleModule {
...
+      Future.wait(_childModules.map((c) => c.resume())).then((_) {
+        onResume().then((_) {
+          _didResumeController.add(this);
+          _isSuspended = false;
+          completer.complete();
+        });
+      });
+    } else {
+      _logger.warning(
+          'Module "$name" is not suspended or not loaded, cannot resume.');
```
trentgrover-wf: same here

georgelesica-wf:
@trentgrover-wf good call, done.

trentgrover-wf:
+1

maxwellpeterson-wf:
+1

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI
 - example works as expected
- [ ] All unit tests pass

Merging into master.

teresarevious-wf:
RM +1

PR #56, opened 2016-09-22
CP-2683 Enable strong-mode/linter. Use `@protected` annotation.

evanweible-wf:
## Improvement
The `meta` package has a `@protected` annotation that we can leverage to mark certain public class methods as callable only by instances of said class. This will be useful for our `onEvent()` methods that are expected to be overridden by subclasses of `Module` in order to respond to a particular event since they should only be called by the internal module logic and not by consumers of the module.

## Changes
- Add `@protected` annotation to:
  - `loadChildModule()`
  - `onWillLoadChildModule()`
  - `onDidLoadChildModule()`
  - `onWillUnloadChildModule()`
  - `onDidUnloadChildModule()`
  - `onLoad()`
  - `onShouldUnload()`
  - `onUnload()`

> Caveat: When a subclass overrides a `@protected` method, they too will have to add the `@protected` annotation to their override in order for the analyzer to enforce the protected behavior on usages of that class method.
>
> However, this actually has a hidden benefit - you can choose to subclass and override a protected method and intentionally omit the `@protected` annotation to sidestep that rule. This is useful for testing (in fact I used this to update an existing test in this PR).

- **Bonus:** enable strong-mode analysis and fix warnings/hints.
- **Bonus:** enable linter and most lint rules and fix lints.
- **Bonus:** deprecate the `LifecycleModule.name` setter. The purpose of the `name` field was to allow modules a way to identify themselves, but there's no reason that value should ever change or be mutable to consumers of a module.

## Breaking Changes
- Changing the signature of several methods from `Future` to `Future<Null>` is technically a stricter typing since `Future` implies `Future<dynamic>`. This could produce an analyzer error for consumers that extend and override these methods, but I _think_ that may only be if strong mode is enabled.

## Testing
- [ ] CI passes

## Code Review
fya: @trentgrover-wf @maxwellpeterson-wf @dustinlessard-wf @jayudey-wf @sebastianmalysa-wf 

fyi (the `@protected` usage may interest you): @todbachman-wf @georgelesica-wf @aaronstgeorge-wf @bencampbell-wf

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,0 +1,63 @@
+analyzer:
+  strong-mode: true
+
+linter:
+  rules:
+    # Error Rules
+    - avoid_empty_else
+    - comment_references
+    - control_flow_in_finally
+    - empty_statements
+    - hash_and_equals
+    - invariant_booleans
+    - iterable_contains_unrelated_type
+    - list_remove_unrelated_type
+    - literal_only_boolean_expressions
+    - test_types_in_equals
+    - throw_in_finally
+    - unrelated_type_equality_checks
+    - valid_regexps
+
+    # TODO: enable these when addressing memory audit/disposables
+    #- cancel_subscriptions
+    #- close_sinks
+
+    # Style Rules
+    - always_declare_return_types
+    - annotate_overrides
+    - avoid_as
+    - avoid_init_to_null
+    - avoid_return_types_on_setters
+    - await_only_futures
+    - camel_case_types
+    - constant_identifier_names
+    - empty_catches
+    - empty_constructor_bodies
+    - implementation_imports
+    - library_names
+    - library_prefixes
+    - non_constant_identifier_names
+    - one_member_abstracts
+    - only_throw_errors
+    - overridden_fields
+    - package_api_docs
+    - package_prefixed_library_names
+    - prefer_is_not_empty
+    - slash_for_doc_comments
+    - sort_constructors_first
+    - sort_unnamed_constructors_first
+    - super_goes_last
+    - type_annotate_public_apis
+    - type_init_formals
+    - unawaited_futures
+    - unnecessary_brace_in_string_interp
+    - unnecessary_getters_setters
+
+    # TODO
+    #- public_member_api_docs
+
+    # This rule forces overly verbose declaration of variables.
+    #- always_specify_types
+
+    # Pub Rules
+    - package_names
--- a/example/panel/modules/data_load_async_module.dart
+++ b/example/panel/modules/data_load_async_module.dart
@@ -12,22 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-library w_module.example.panel.modules.dataLoadAsync_module;
+library w_module.example.panel.modules.data_load_async_module;
 
 import 'dart:async';
 
-import 'package:w_flux/w_flux.dart';
+import 'package:meta/meta.dart' show protected;
 import 'package:react/react.dart' as react;
+import 'package:w_flux/w_flux.dart';
 import 'package:w_module/w_module.dart';
 
 class DataLoadAsyncModule extends Module {
+  @override
   final String name = 'DataLoadAsyncModule';
 
   DataLoadAsyncActions _actions;
-  DataLoadAsyncStore _stores;
-
   DataLoadAsyncComponents _components;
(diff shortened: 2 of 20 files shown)


aviary-wf:
## Raven
Number of Findings: 0

rmconsole-wf:
@evanweible-wf
* It looks like you have not included any tickets in the title of this PR, but this repo requires tickets for all pull requests. Please create a ticket for this change, or click <a href='https://w-rmconsole.appspot.com/api/v1/pulls/86378731/addTicketToTitle'>here</a> to have Rosie create one for you.

georgelesica-wf:
+1 lgtm

maxwellpeterson-wf:
+1

trentgrover-wf:
https://github.com/Workiva/w_module/pull/55 has merged so those methods can be incorporated now

evanweible-wf:
@trentgrover-wf @maxwellpeterson-wf @georgelesica-wf rebased against latest and included the `@protected` annotation to the new methods from #55. 

I also noticed that with strong mode, changing `Future` to `Future<Null>` for return types could produce an issue for consumers that extend and override - they would need to update their signatures as well. I can get a confirmation on exactly when this is an issue and what level of severity the analyzer messages is if we think this is a sticking point.

analysis_options.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - iterable_contains_unrelated_type
+    - list_remove_unrelated_type
+    - literal_only_boolean_expressions
+    - test_types_in_equals
+    - throw_in_finally
+    - unrelated_type_equality_checks
+    - valid_regexps
+
+    # TODO: enable these when addressing memory audit/disposables
+    #- cancel_subscriptions
```
dustinlessard-wf: This is awesome.

analysis_options.yaml
```diff
@@ -0,0 +1,63 @@
+analyzer:
+  strong-mode: true
+
+linter:
+  rules:
```
dustinlessard-wf: Do we have a recommended list of which rules to use and if so, is this it?

example/panel/modules/data_load_async_module.dart
```diff
@@ -12,32 +12,38 @@
...
     _stores = new DataLoadAsyncStore(_actions);
     _components = new DataLoadAsyncComponents(_actions, _stores);
   }
 
-  Future onLoad() async {
+  @override
+  DataLoadAsyncComponents get components => _components;
+
+  @override
+  @protected
```
dustinlessard-wf: Are we aware of any additional overhead with using the 'meta' package or is it only at play prior to transpilation?

analysis_options.yaml
georgelesica-wf: Here's the list we used in another library, maybe we should discuss and come up with an informal standard at least...

```
linter:
  rules:
    # Errors
    - avoid_empty_else
    - hash_and_equals
    - iterable_contains_unrelated_type
    - test_types_in_equals
    - throw_in_finally
    - unrelated_type_equality_checks
    # Style
    - always_declare_return_types
    - annotate_overrides
    - avoid_init_to_null
    - avoid_return_types_on_setters
    - await_only_futures
    - camel_case_types
    - constant_identifier_names
    - empty_constructor_bodies
    - implementation_imports
    - library_names
    - library_prefixes
    - non_constant_identifier_names
    - one_member_abstracts
    - overridden_fields
    - package_api_docs
    - package_prefixed_library_names
    - prefer_is_not_empty
    - slash_for_doc_comments
    - sort_constructors_first
    - sort_unnamed_constructors_first
    - super_goes_last
    - type_init_formals
    - unnecessary_getters_setters
    # Pub rules
    - package_names
```

example/panel/modules/data_load_async_module.dart
evanweible-wf: I don't think so - as far as I can tell, they're just tools for the analyzer to leverage and have effect on the built JS.

maxwellpeterson-wf:
+1

georgelesica-wf:
+1

trentgrover-wf:
+1
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI
- [ ] All unit tests pass

Merging into master.

teresarevious-wf:
RM +1

PR #59, opened 2016-11-07
MOBILE-2277: Implement serializable w_module

brianblanchard-wf:
## Description
- Added a way to serialize w_modules to allow for communication with other systems (such as a mobile app)

## What was changed
- A separate entry point for a serializable module so all the existing w_module logic does not have any browser specific dependencies.
- Added a `SerializableBus` singleton that registers for events and listens for api calls to modules
- Added a `SerializableModule` class that extends the existing `Module` class that adds needed information to serialize a module
- Added a `Reflectable` class that allows classes and methods to be reflected upon using mirrors.
- Added a `SerializableEvent` and `SerializableEvents` class that allows events to be serializable.
- Unit tests around additions and changes.

Please review: @trentgrover-wf @evanweible-wf @maxwellpeterson-wf 

An actual implementation of these changes can be seen in this [diff](https://github.com/Workiva/w_viewer/compare/master...brianblanchard-wf:MOBILE-2281)

### DIFF
--- a/lib/src/serializable.dart
+++ b/lib/src/serializable.dart
@@ -19,10 +19,11 @@ import 'dart:mirrors';
 import 'dart:async';
 
 import 'package:logging/logging.dart';
-import 'package:w_common/w_common.dart' show JsonSerializable, Disposable;
+import 'package:w_common/disposable.dart' show Disposable;
+import 'package:w_common/json_serializable.dart' show JsonSerializable;
 
-import 'event.dart';
-import 'module.dart';
+import 'package:w_module/src/event.dart';
+import 'package:w_module/src/module.dart';
 
 // Any classes / methods that are going to be reflected must annotate with this
 class Reflectable {
@@ -80,13 +81,14 @@ class SerializableBus {
       <String, _ModuleRegistration>{};
 
   Bridge get bridge => _bridge;
-  Map<String, _ModuleRegistration> get moduleRegistrations =>
-      _moduleRegistrations;
-
   set bridge(Bridge bridge) {
     _bridge = bridge;
     _bridge.apiCallReceived.listen(_handleApiCall);
   }
+
+  Map<String, SerializableModule> get registeredModules =>
+      new Map.fromIterable(_moduleRegistrations.keys,
+          value: (key) => _moduleRegistrations[key].module);
 
   void reset() {
     _moduleRegistrations.clear();
@@ -118,7 +120,7 @@ class SerializableBus {
 
     SerializableModule targetModule = _moduleRegistrations[module].module;
 
-    if (targetModule != null) {
+    if (targetModule != null && data is List) {
       _deserializeAndCall(targetModule, method, data);
     }
   }
@@ -127,74 +129,72 @@ class SerializableBus {
       SerializableModule module, String method, List data) {
     InstanceMirror apiMirror = reflect(module.api);
 
-    if (apiMirror != null) {
-      ClassMirror classMirror = apiMirror.type;
+    if (apiMirror == null) {
+      _logger.warning(
+          'Unable to create mirror on api for ${module.serializableKey}');
+      return;
+    }
 
-      if (ClassMirror != null) {
-        MethodMirror apiMethodMirror =
-            classMirror.declarations[new Symbol(method)];
+    ClassMirror classMirror = apiMirror.type;
+    MethodMirror apiMethodMirror = classMirror.declarations[new Symbol(method)];
 
-        if (apiMethodMirror == null) {
-          _logger.warning(
-              'Method $method does not exist on ${module.serializableKey}\' module\'s API');
-          return;
-        }
+    if (apiMethodMirror == null) {
+      _logger.warning(
+          'Method $method does not exist on ${module.serializableKey}\' module\'s API');
+      return;
+    }
 
-        // Check here that the position args in data match the expected params of the method being called
-        if (apiMethodMirror.parameters.length == data.length) {
-          for (var i = 0; i < apiMethodMirror.parameters.length; i++) {
-            var param = apiMethodMirror.parameters[i];
+    // Check here that the position args in data match the expected params of the method being called
+    if (apiMethodMirror.parameters.length == data.length) {
+      for (var i = 0; i < apiMethodMirror.parameters.length; i++) {
+        var param = apiMethodMirror.parameters[i];
 
-            if (data[i] is Map && param.type.reflectedType != Map) {
-              ClassMirror paramClassMirror =
-                  reflectClass(param.type.reflectedType);
+        if (data[i] is Map && param.type.reflectedType != Map) {
+          ClassMirror paramClassMirror = reflectClass(param.type.reflectedType);
 
-              // Paramter type must implement fromJson name constructor that takes a Map
-              try {
-                var instance = paramClassMirror
-                    .newInstance(new Symbol('fromJson'), [data[i]]).reflectee;
-                data[i] = instance;
-              } on NoSuchMethodError {
-                _logger.warning(
-                    '${paramClassMirror.simpleName.toString()} does not implement fromJson named constructor');
-                return;
-              }
-            }
+          // Paramter type must implement fromJson name constructor that takes a Map
+          try {
+            var instance = paramClassMirror
+                .newInstance(new Symbol('fromJson'), [data[i]]).reflectee;
+            data[i] = instance;
+          } on NoSuchMethodError {
+            _logger.warning(
+                '${paramClassMirror.simpleName.toString()} does not implement fromJson named constructor');
+            return;
           }
-
-          if (apiMirror != null) {
-            try {
-              apiMirror.invoke(new Symbol(method), data);
-            } catch (e) {
-              _logger.severe(
-                  'Unable to call $method on ${module.serializableKey} w_module, ${e.toString()}');
-            }
-          }
-        } else {
-          _logger.warning(
-              'Unable to call api method $method in ${module.serializableKey} w_module, mismatched params');
         }
       }
+
+      try {
+        apiMirror.invoke(new Symbol(method), data);
+      } catch (e) {
+        _logger.severe(
+            'Unable to call $method on ${module.serializableKey} w_module, ${e.toString()}');
+      }
+    } else {
+      _logger.warning(
+          'Unable to call api method $method in ${module.serializableKey} w_module, mismatched params');
     }
   }
 
-  void _registerForAllEvents(SerializableModule module) {
-    if (module.events != null) {
-      for (var event in module.events.allEvents) {
+  void _registerForAllEvents(_ModuleRegistration registration) {
+    if (registration.module.events != null) {
+      for (var event in registration.module.events.allEvents) {
         if (event is SerializableEvent) {
-          event
-              .listen((payload) => _sendEvent(module, event.eventKey, payload));
+          registration.manageStreamSubscription(event.listen((payload) =>
+              _sendEvent(registration.module, event.eventKey, payload)));
         }
       }
     } else {
-      _logger.info('Events not defined for ${module.serializableKey} module');
+      _logger.info(
+          'Events not defined for ${registration.module.serializableKey} module');
     }
   }
 
   void _registerForLifecycleEvents(_ModuleRegistration registration) {
     registration
         .manageStreamSubscription(registration.module.willLoad.listen((_) {
-      _registerForAllEvents(registration.module);
+      _registerForAllEvents(registration);
       _sendEvent(registration.module, 'willLoad', null);
     }));
     registration.manageStreamSubscription(registration.module.didLoad
--- a/lib/src/serializable.dart
+++ b/lib/src/serializable.dart
@@ -202,7 +202,7 @@ class SerializableBus {
             ClassMirror paramClassMirror =
                 reflectClass(param.type.reflectedType);
 
-            // Paramter type must implement fromJson model that takes a Map
+            // Paramter type must implement fromJson name constructor that takes a Map
             try {
               var instance = paramClassMirror
                   .newInstance(new Symbol('fromJson'), [data[i]]).reflectee;
--- a/lib/src/serializable.dart
+++ b/lib/src/serializable.dart
@@ -17,11 +17,9 @@ library serializable_module.src.serializable;
 @MirrorsUsed(metaTargets: 'serializable_module.src.serializable.Reflectable')
 import 'dart:mirrors';
 import 'dart:async';
-import 'dart:html' show window, CustomEvent;
-import 'dart:js' show JsObject, context;
 
 import 'package:logging/logging.dart';
-import 'package:w_common/w_common.dart' show JsonSerializable;
+import 'package:w_common/w_common.dart' show JsonSerializable, Disposable;
 
 import 'event.dart';
 import 'module.dart';
@@ -32,61 +30,27 @@ class Reflectable {
 }
 
 abstract class Bridge<T> {
-  StreamController<Map> _eventController;
-
-  Bridge() {
-    _eventController = new StreamController<Map>.broadcast();
-  }
-
-  void broadcast(Map dataToSend);
-  void handleEvent(T apiCall);
-
-  Stream<Map> get eventReceived => _eventController.stream;
-}
-
-// This class is used to communicate across a WKWebView in an iOS app
-class WKWebViewBridge extends Bridge<CustomEvent> {
-  WKWebViewBridge() {
-    window.on['bridge'].listen(handleEvent);
-  }
-
-  @override
-  void broadcast(Map dataToSend) {
-    wkBridge?.callMethod('postMessage', [new JsObject.jsify(dataToSend)]);
-  }
-
-  @override
-  void handleEvent(CustomEvent apiCall) {
-    _eventController.add(apiCall.detail);
-  }
-
-  JsObject get wkBridge {
-    if (context['webkit'] != null) {
-      return context['webkit']['messageHandlers']['bridge'];
-    } else {
-      return null;
-    }
-  }
+  final Stream<Map> apiCallReceived;
+  Bridge(this.apiCallReceived);
+  void broadcastSerializedEvent(Map event);
+  void handleSerializedApiCall(T apiCall);
 }
 
 class SerializableEvent<T> extends Event<T> {
+  String _eventKey;
+
   SerializableEvent(this._eventKey, DispatchKey dispatchKey)
       : super(dispatchKey);
 
-  String _eventKey;
   String get eventKey => _eventKey;
 }
 
 abstract class SerializableEvents {
-  List<SerializableEvent> get allEvents => [];
+  List<SerializableEvent> get allEvents;
 }
 
 abstract class SerializableModule extends Module {
   SerializableModule() {
-    _registerWithSerializableBus();
-  }
-
-  void _registerWithSerializableBus() {
     SerializableBus.sharedBus.registerModule(this);
   }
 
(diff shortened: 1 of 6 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+    if (module.events != null) {
+      for (var event in module.events.allEvents) {
+        if (event is SerializableEvent) {
+          event
+              .listen((payload) => _sendEvent(module, event.eventKey, payload));
+        }
+      }
+    } else {
+      _logger
+          .warning('Events not defined for ${module.serializableKey} module');
```
evanweible-wf: I think `info` would be more appropriate here - modules aren't required to define any events.

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+    });
+  }
+
+  void _sendEvent(SerializableModule module, String eventKey, Object data) {
+    Map dataToSend = new Map();
+    dataToSend['module'] = module.serializableKey;
+    dataToSend['event'] = eventKey;
+
+    if (data is JsonSerializable) {
+      data = (data as JsonSerializable).toJson();
```
evanweible-wf: You should be able to avoid the explicit `as` type cast here since you know it is of type `JsonSerializable` via the `if` check.

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+
+    _bridge?.broadcast(dataToSend);
+  }
+
+  void _handleApiCall(Map apiCall) {
+    String module = apiCall['module'];
+    String method = apiCall['method'];
+    Object data = apiCall["data"];
+
+    SerializableModule targetModule = _registeredModules[module];
```
evanweible-wf: null check this

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+
+    _deserializeAndCall(targetModule, method, data);
+  }
+
+  void _deserializeAndCall(
+      SerializableModule module, String method, List data) {
+    InstanceMirror apiMirror = reflect(module.api);
+    ClassMirror classMirror = apiMirror.type;
```
evanweible-wf: null check on `apiMirror`

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+    MethodMirror apiMethodMirror = classMirror.declarations[new Symbol(method)];
```
evanweible-wf: null check on `classMirror`

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+
+    if (apiMethodMirror == null) {
+      _logger.warning(
+          'Method $method does not exist on ${module.serializableKey}\' module\'s API');
+      return;
+    }
+
+    // Check here that the position args in data match the expected params of the method being called
+    if (data is List && apiMethodMirror.parameters.length == data.length) {
```
evanweible-wf: The method signature already requires that `data` is a `List`, so this first part should never be false

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+      for (var i = 0; i < apiMethodMirror.parameters.length; i++) {
+        var param = apiMethodMirror.parameters[i];
+
+        // If the type data being passed to this param is not equal to the expected type
+        // try to serialize it into an Dart class
+        if (param.type.reflectedType != data[i].runtimeType) {
```
evanweible-wf: I'm not very familiar with mirrors at all - how does this handle generics? For example, if the reflected param type is `List<int>` but the data runtime type could only be inferred to be `List` but in practice only contained `int`s, would that be an issue?

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+              var instance = paramClassMirror
+                  .newInstance(new Symbol('fromJson'), [data[i]]).reflectee;
+              data[i] = instance;
+            } on NoSuchMethodError {
+              _logger.warning(
+                  '${paramClassMirror.simpleName.toString()} does not implement fromJson named constructor');
+              return;
+            }
+          } else {
+            _logger.warning('Incompatiable type for deserialization');
```
evanweible-wf: Maybe add some more context here (module key, event key, expected type vs. actual incompatible type)

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  void broadcast(Map dataToSend);
+  void handleEvent(T apiCall);
+
+  Stream<Map> get eventReceived => _eventController.stream;
+}
+
+// This class is used to communicate across a WKWebView in an iOS app
+class WKWebViewBridge extends Bridge<CustomEvent> {
```
brianblanchard-wf: Where do you think this should live? I was told to keep it in here for now.

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+
+import 'package:logging/logging.dart';
+import 'package:w_common/w_common.dart' show JsonSerializable;
+
+import 'event.dart';
+import 'module.dart';
+
+// Any classes / methods that are going to be reflected must annotate with this
+class Reflectable {
+  const Reflectable();
```
brianblanchard-wf: No you have to define your own class you annotate with. https://api.dartlang.org/dev/1.20.0-dev.9.0/dart-mirrors/MirrorsUsed/metaTargets.html

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  @override
+  void handleEvent(CustomEvent apiCall) {
+    _eventController.add(apiCall.detail);
+  }
+
+  JsObject get wkBridge {
+    if (context['webkit'] != null) {
+      return context['webkit']['messageHandlers']['bridge'];
```
brianblanchard-wf: If the `webkit` is there then the `messageHandlers` will always be there. Also, when the `wkBridge` is invoked using an optional `wkBridge?.callMethod('postMessage', [new JsObject.jsify(dataToSend)]); `

lib/src/serializable.dart
```diff
@@ -0,0 +1,237 @@
...
+    }
+  }
+}
+
+class SerializableEvent<T> extends Event<T> {
+  SerializableEvent(this._eventKey, DispatchKey dispatchKey)
+      : super(dispatchKey);
+
+  String _eventKey;
+  String get eventKey => _eventKey;
```
brianblanchard-wf: That is a problem we will have to tackle in the future, being able to register a unique instance with bridge. When I talked with Trent initially we decided for now it is ok to just be able to handle one instance of a given module at a time.

lib/src/serializable.dart
evanweible-wf: I think it can live with the application you're working on until it has been vetted and at that point it could probably live in wdesk_sdk or something similar

lib/src/serializable.dart
brianblanchard-wf: Without the explicit cast, there is an analyzer error.

evanweible-wf: Okay, it's fine as is then. Out of curiosity - what version of the Dart SDK are you on?

brianblanchard-wf: 1.19.1

lib/src/serializable.dart
brianblanchard-wf: It will work. After some testing I had to make some changes to how this works. That `if (param.type.reflectedType != data[i].runtimeType` check does not alway work as expected. For example, when passing an array from JS, the type is `JsArray` but if you do a type check on like `jsArray is List` that returns true. So now it checks to see if data being passed is a `Map` and then will try to serialize it into a the expected object of the api method.

I know this is not idea, but before this is used at any type of scale, all this serialization stuff will be replaced by frugal so I think it is ok for now.

brianblanchard-wf:
@evanweible-wf thanks for all the great feedback. I believe I addressed all your concerns. Please take another look when you have time.

lib/src/serializable.dart
evanweible-wf: Yup, no problem, was mainly just curious

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+  static final SerializableBus sharedBus = new SerializableBus();
+
+  Bridge _bridge;
+  final Logger _logger = new Logger('Serializable Bus');
+  Map<String, _ModuleRegistration> _moduleRegistrations =
+      <String, _ModuleRegistration>{};
+
+  Bridge get bridge => _bridge;
+  Map<String, _ModuleRegistration> get moduleRegistrations =>
+      _moduleRegistrations;
```
evanweible-wf: I'm not sure it makes sense to expose the `_ModuleRegistration` instances, I see them as being implementation details that make disposal easier. It think exposing the modules like you had before would be fine:

```dart
Map<String, SerializableModule> get registeredModules
    => new Map.fromIterable(_moduleRegistrations.keys, value: (key) => _moduleRegistrations[key].module);
```

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+
+  set bridge(Bridge bridge) {
```
evanweible-wf: If a field has a getter and a setter, they should be grouped together

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+      ClassMirror classMirror = apiMirror.type;
+
+      if (ClassMirror != null) {
+        MethodMirror apiMethodMirror =
+            classMirror.declarations[new Symbol(method)];
+
+        if (apiMethodMirror == null) {
+          _logger.warning(
+              'Method $method does not exist on ${module.serializableKey}\' module\'s API');
+          return;
```
evanweible-wf: I like this approach of logging/returning early in the failure case instead of nested ifs, could you do this for the `apiMirror` and `classMirror` null checks above, too?

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  void _deserializeAndCall(
+      SerializableModule module, String method, List data) {
+    InstanceMirror apiMirror = reflect(module.api);
+
+    if (apiMirror != null) {
+      ClassMirror classMirror = apiMirror.type;
+
+      if (ClassMirror != null) {
```
evanweible-wf: s/`ClassMirror`/`classMirror`

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+          }
+        } else {
+          _logger.warning(
+              'Unable to call api method $method in ${module.serializableKey} w_module, mismatched params');
+        }
+      }
+    }
+  }
+
+  void _registerForAllEvents(SerializableModule module) {
```
evanweible-wf: you'll want to pass in the `_ModuleRegistration` instance here, too, so that you can have it manage the event subscriptions below

test/serializable_test.dart
```diff
@@ -0,0 +1,333 @@
...
+
+      when(bridge.apiCallReceived).thenReturn(bridgeEventController.stream);
+
+      bus.registerModule(module);
+      bus.bridge = bridge;
+    });
+
+    test('should provide a singleton', () async {
+      expect(SerializableBus.sharedBus, isNotNull);
+      expect(SerializableBus.sharedBus is SerializableBus, isTrue);
```
evanweible-wf:
```dart
expect(SerializableBus.sharedBus, new isInstanceOf<SerializableBus>());
```

tool/dev.dart
```diff
@@ -31,6 +31,7 @@ Future<Null> main(List<String> args) async {
   ];
   config.copyLicense.directories = directories;
   config.format.directories = directories;
+  config.test.platforms = ['content-shell', 'vm'];
```
evanweible-wf: If you switch to the individual w_common entry points above, I believe you'll be able to run these tests on the VM instead of in the browser

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+      SerializableModule module, String method, List data) {
+    InstanceMirror apiMirror = reflect(module.api);
+
+    if (apiMirror == null) {
+      _logger.warning(
+          'Unable to create mirror on api for ${module.serializableKey}');
+      return;
+    }
+
+    ClassMirror classMirror = apiMirror.type;
```
brianblanchard-wf: A null check not needed here. If the `apiMirror` exists then a class mirror must also exist.

brianblanchard-wf:
@evanweible-wf I have addressed your last round of feedback. Please review.

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+library serializable_module.src.serializable;
+
+@MirrorsUsed(metaTargets: 'serializable_module.src.serializable.Reflectable')
+import 'dart:mirrors';
+import 'dart:async';
```
evanweible-wf: #nit alphabetize imports

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+  void handleSerializedApiCall(T apiCall);
+}
+
+class SerializableEvent<T> extends Event<T> {
+  String _eventKey;
+
+  SerializableEvent(this._eventKey, DispatchKey dispatchKey)
+      : super(dispatchKey);
+
+  String get eventKey => _eventKey;
```
evanweible-wf: could simplify this to a final variable:

```dart
final String eventKey;
SerializableEvent(this.eventKey, DispatchKey dispatchKey) : super(dispatchKey);
```

evanweible-wf: could simplify this to a final variable:

```dart
final String eventKey;
SerializableEvent(this.eventKey, DispatchKey dispatchKey) : super(dispatchKey);
```

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+    MethodMirror apiMethodMirror = classMirror.declarations[new Symbol(method)];
+
+    if (apiMethodMirror == null) {
+      _logger.warning(
+          'Method $method does not exist on ${module.serializableKey}\' module\'s API');
+      return;
+    }
+
+    // Check here that the position args in data match the expected params of the method being called
+    if (apiMethodMirror.parameters.length == data.length) {
```
evanweible-wf: could also invert this if check here and return early if `apiMethodMirror.parameters.length != data.length`

test/serializable_test.dart
```diff
@@ -0,0 +1,317 @@
...
+@TestOn('vm || browser')
+library serializable_module.test.serializable_test;
+
+import 'dart:async';
+
+import 'package:w_module/serializable_module.dart';
+import 'package:w_common/json_serializable.dart' show JsonSerializable;
+import 'package:w_module/w_module.dart';
+import 'package:test/test.dart';
+import 'package:mockito/mockito.dart';
```
evanweible-wf: sort imports

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+
+  Bridge _bridge;
+  final Logger _logger = new Logger('Serializable Bus');
+  Map<String, _ModuleRegistration> _moduleRegistrations =
+      <String, _ModuleRegistration>{};
+
+  Bridge get bridge => _bridge;
+  set bridge(Bridge bridge) {
+    _bridge = bridge;
+    _bridge.apiCallReceived.listen(_handleApiCall);
```
evanweible-wf: Should store this `StreamSubscription` and then cancel it when a new bridge is set

evanweible-wf:
@Workiva/client-platform-pp @Workiva/ref-pp anyone else care to take a look?

dustinlessard-wf:
Looking - will post when done.

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  Map<String, SerializableModule> get registeredModules =>
+      new Map.fromIterable(_moduleRegistrations.keys,
+          value: (key) => _moduleRegistrations[key].module);
+
+  void reset() {
+    _moduleRegistrations.clear();
+    _bridge = null;
```
dustinlessard-wf: Looking at the [Bridge abstract class](https://github.com/Workiva/w_module/pull/59/files#diff-949ea1a52f226534b3058b4ce26a64edR33), and the [proposed implementation](https://github.com/Workiva/w_viewer_mobile/blob/MOBILE-2307/lib/src/wk_web_view_bridge.dart#L17) : 
We see that there is a `StreamSubscription` created yet never cancelled.

Looking at the reset method, we see that we simply null out the reference to _bridge.  

Two questions: 
1. Should `WKWebViewBridge` support a function to call when we want to clean up resources (aka dispose)?
2. Should `SerializableBus` call that cleanup function prior to nulling out the `_bridge` reference?

One option is to consider using w_common : https://github.com/Workiva/w_common/blob/master/lib/disposable.dart

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+    event['module'] = module.serializableKey;
+    event['event'] = eventKey;
+
+    if (data is JsonSerializable) {
+      data = (data as JsonSerializable).toJson();
+    }
+
+    event['data'] = data;
+
+    _bridge?.broadcastSerializedEvent(event);
```
dustinlessard-wf: Do we want to know if the `_bridge` is null at the time that we attempt to send an event?  If so, we may consider adding some logging around that.

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+  void _registerForAllEvents(_ModuleRegistration registration) {
+    if (registration.module.events != null) {
+      for (var event in registration.module.events.allEvents) {
+        if (event is SerializableEvent) {
+          registration.manageStreamSubscription(event.listen((payload) =>
+              _sendEvent(registration.module, event.eventKey, payload)));
+        }
+      }
+    } else {
+      _logger.info(
```
brianblanchard-wf: In an earlier comment @evanweible-wf suggested that `info` were to be used because what it is logging is not a required for a module to implement.

lib/src/serializable.dart
```diff
@@ -0,0 +1,224 @@
...
+
+          // Paramter type must implement fromJson name constructor that takes a Map
+          try {
+            var instance = paramClassMirror
+                .newInstance(new Symbol('fromJson'), [data[i]]).reflectee;
+            data[i] = instance;
+          } on NoSuchMethodError {
+            _logger.warning(
+                '${paramClassMirror.simpleName.toString()} does not implement fromJson named constructor');
+            return;
```
dustinlessard-wf: nit: Is this needed?

brianblanchard-wf: I think it is useful because we get a meaningful error message instead of the error being thrown in the console.

brianblanchard-wf:
@dustinlessard-wf @evanweible-wf I have addressed for latest feedback. I added the `Disposable` mixin to the `Bridge` abstract class that is now used to manage the stream subscriptions to the bridge.

lib/src/serializable.dart
```diff
@@ -0,0 +1,235 @@
...
+  static final SerializableBus sharedBus = new SerializableBus();
+
+  Bridge _bridge;
+  final Logger _logger = new Logger('Serializable Bus');
+  Map<String, _ModuleRegistration> _moduleRegistrations =
+      <String, _ModuleRegistration>{};
+
+  Bridge get bridge => _bridge;
+  set bridge(Bridge bridge) {
+    _bridge = bridge;
```
dustinlessard-wf: Caveat: I'm uncertain of the answer to this question:
If `_bridge` is not null at the time we attempt to set it, should we `dispose` the previous one?

brianblanchard-wf: I don't think it would hurt to do that.

lib/src/serializable.dart
```diff
@@ -30,20 +30,23 @@ class Reflectable {
...
 
-abstract class Bridge<T> {
+abstract class Bridge<T> extends Object with Disposable {
   final Stream<Map> apiCallReceived;
   Bridge(this.apiCallReceived);
   void broadcastSerializedEvent(Map event);
   void handleSerializedApiCall(T apiCall);
+
+  @override
+  void manageStreamSubscription(StreamSubscription subscription) {
```
evanweible-wf: I think this should be private - by overriding it and omitting the `@protected` label, you're effectively allowing anyone who has access to the bridge to add subscriptions that should be managed by that bridge, when in reality that should only be a handled by the bridge itself. If you make this private, then you can still call it below without exposing that ability to external consumers.

lib/src/serializable.dart
```diff
@@ -82,8 +85,13 @@ class SerializableBus {
 
   Bridge get bridge => _bridge;
   set bridge(Bridge bridge) {
+    if (_bridge != null) {
+      _bridge.dispose();
+    }
```
evanweible-wf: `_bridge?.dispose()`

lib/src/serializable.dart
```diff
@@ -82,8 +85,13 @@ class SerializableBus {
...
+
     _bridge = bridge;
-    _bridge.apiCallReceived.listen(_handleApiCall);
+    _bridge.manageStreamSubscription(
```
evanweible-wf: Probably need to ensure that `bridge` isn't being set to null here, otherwise this would throw (would also prevent a failure below when reset() is called)

lib/src/serializable.dart
```diff
@@ -0,0 +1,239 @@
...
+
+abstract class Bridge<T> extends Object with Disposable {
+  final Stream<Map> apiCallReceived;
+  Bridge(this.apiCallReceived);
+  void broadcastSerializedEvent(Map event);
+  void handleSerializedApiCall(T apiCall);
+
+  @override
+  void manageStreamSubscription(StreamSubscription subscription) {
+    super.manageStreamSubscription(subscription);
```
maxwellpeterson-wf: Does this method need to be overridden if it is just calling super?

lib/src/serializable.dart
```diff
@@ -0,0 +1,239 @@
...
+}
+
+class _ModuleRegistration extends Object with Disposable {
+  final SerializableModule module;
+
+  _ModuleRegistration(this.module);
+
+  @override
+  void manageStreamSubscription(StreamSubscription subscription) {
+    super.manageStreamSubscription(subscription);
```
maxwellpeterson-wf: Same question about overriding here.

lib/src/serializable.dart
evanweible-wf: This should probably be private and only used in this file, otherwise external consumers could tell the bridge to manage stream subscriptions that it shouldn't

evanweible-wf:
+1

dustinlessard-wf:
Are we ok with decreasing coverage by nearly 7%?  Is that accurate?

brianblanchard-wf:
@dustinlessard-wf I have added some more unit tests. Is the level of coverage now acceptable?

dustinlessard-wf:
+1 - I think it was probably acceptable before, but I wasn't certain about how accurate CodeCov was being.  (I still have my doubts but this is totally great!)

evanweible-wf:
+1

sebastianmalysa-wf:
+1

maxwellpeterson-wf:
+1

jordanross-wf:
+10 from a functionality standpoint

maxwellpeterson-wf:
@jayudey-wf ready for you

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction (from offline discussion with brian)
- [ ] Dev +1's
 - performed and documented by Jordan
- [ ] All unit tests pass

Merging into master.

PR #62, opened 2016-12-21
RAP-1482 Mix Disposable into LifecycleModule

todbachman-wf:
**Jira Ticket:** [RAP-1482](https://jira.atl.workiva.net/browse/RAP-1482)

## Problem:
We'd like to use `Disposable` in the `LifecycleModule`. This will allow us to properly close the stream controllers that it manages.

## Solution:
1. Mixed in `Disposable`.
1. Aliased `dispose` to `unload` and marked it `@deprecated`. This will help us guide consumers to rely on `unload` instead.
1. Overrode `onDispose` and documented it as deprecated as well.

**Note:** there will be at least two more PRs to follow: one to enforce a single load/unload cycle; and another to make load, unload, suspend, and resume noops when the module is already in the correct state.

## Semantic Versioning:

- [ ] **Patch**
  - [ ] This change does not affect the public API
  - [ ] This change fixes existing incorrect behavior without any additions
- [x] **Minor**
  - [x] This change adds something to the public API
  - [ ] This change deprecates a part of the public API
* [ ] Major
  - [ ] This change modifies part of the public API in a backwards-incompatible manner
  - [ ] This change removes part of the public API

## +10/QA:
- [ ] CI build passes

## Review:
@Workiva/rich-app-platform-pp 
@Workiva/web-platform-pp

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -18,10 +18,11 @@ import 'dart:async';
 
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart' show protected;
+import 'package:w_common/disposable.dart';
 
 /// Intended to be extended by most base module classes in order to provide a
 /// unified lifecycle API.
-abstract class LifecycleModule {
+abstract class LifecycleModule extends Object with Disposable {
   bool _isLoaded = false;
 
   bool _isSuspended = false;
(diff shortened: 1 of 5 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

test/event_test.dart
```diff
@@ -53,10 +53,10 @@ void main() {
       // The point of this test is to exercise the `where` method which is made available
       // on an action by extending stream and overriding `listen`
       Stream<String> filteredStream = event.where((value) => value == 'water');
-      expectAsync(filteredStream.listen)((payload) {
+      filteredStream.listen(expectAsync1((payload) {
```
todbachman-wf: This fixes a deprecation warning.

lib/src/serializable.dart
```diff
@@ -225,7 +225,8 @@ class SerializableBus {
     event['event'] = eventKey;
 
     if (data is JsonSerializable) {
-      data = (data as JsonSerializable).toJson();
+      JsonSerializable serializeData = data;
+      data = serializeData.toJson();
```
bencampbell-wf: There may be more type safety changes to be had with this class. 😰

lib/src/lifecycle_module.dart
```diff
@@ -297,6 +303,13 @@ abstract class LifecycleModule {
     return completer.future;
   }
 
+  /// Aliased to [unload].
+  ///
+  /// Deprecated: Use the method [unload] instead.
+  @deprecated
```
maxwellpeterson-wf: Can we use `@protected` here?

lib/src/lifecycle_module.dart
```diff
@@ -297,6 +303,13 @@ abstract class LifecycleModule {
...
+  @override
+  Future<Null> dispose() => unload();
```
maxwellpeterson-wf: Can we use `@protected()` here?

maxwellpeterson-wf:
+1

lib/src/lifecycle_module.dart
todbachman-wf: @maxwellpeterson-wf would you like me to add it in addition to the other attributes?

maxwellpeterson-wf: Discussed in chat. `@protected` doesn't prevent consumers from calling `dispose()` in their subclass of `LifecycleModule`. This is fine.

todbachman-wf: We decided to use `@deprecated` instead of `@protected` because it would "protect" us from consumers calling `dispose` from within their modules as well as from outside.

bencampbell-wf:
+10

Things Tested
-------------
- [x] CI passed

todbachman-wf:
@jayudey-wf -- Looks like RAP doesn't have permission to merge to this repo. 😢 So, this is ready for you.

trentgrover-wf:
+1
merging - FYI @jayudey-wf

PR #64, opened 2016-12-23
RAP-1484 Enforce LifecycleModule State Transitions

todbachman-wf:
**Jira Ticket:** [RAP-1484](https://jira.atl.workiva.net/browse/RAP-1484)

## Problem:
We should enforce proper transitions between `LifecycleModule` states. The arrows in the following diagram identify valid state transitions. Green arrows represent transitions that will proceed as requested. Blue arrows represent noops.

lifecycle module - lifecycle

## Solution:
Updated `LifecycleModule` to enforce the valid transitions. If a transition will be a noop, log a warning. If an invalid transition is requested a `StateError` is thrown.

## +10/QA:
- [ ] CI build passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,14 @@ Method         | Description
 `shouldUnload` | Returns the unloadable state of the `Module` and its child modules as a `ShouldUnloadResult`.  Internally, this executes the module's `onShouldUnload` method.
 `unload`       | Triggers unloading of a `Module` and all of its child modules.  Internally, this executes the module's `shouldUnload` method, and, if that completes successfully, executes the module's `onUnload` method. If unloading is rejected, this method will complete with an error.
 
+![lifecycle module - lifecycle](https://cloud.githubusercontent.com/assets/11619752/21526229/d7a5ae1c-ccdf-11e6-9e95-ff16a83e3a7f.png)
+
+The graphic above illustrates legal lifecycle state transitions. Any state
+transition that is not defined will throw a `StateError`. No-op transitions are
+illustrated with blue arrows. Calling the lifecycle method when the module is
+already in the given state will result in a logger warning. If a no-op is
+performed on a transitioning state the pending transition future is returned.
+
 ### Lifecycle Events
 
 `Module` also exposes lifecycle event streams that an external consumer can listen to:
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -17,13 +17,13 @@ library w_module.src.lifecycle_module;
 import 'dart:async';
 
 import 'package:logging/logging.dart';
-import 'package:meta/meta.dart' show protected;
+import 'package:meta/meta.dart' show protected, required;
 import 'package:w_common/disposable.dart';
 
 /// Possible states a [LifecycleModule] may occupy.
 enum LifecycleState {
   /// The module has been instantiated.
-  initialized,
+  instantiated,
 
   /// The module is in the process of being loaded.
   loading,
@@ -52,16 +52,20 @@ enum LifecycleState {
 abstract class LifecycleModule extends Object with Disposable {
   Logger _logger;
   String _name = 'Module';
-  LifecycleState _state = LifecycleState.initialized;
+  LifecycleState _state = LifecycleState.instantiated;
   Future<Null> _transitionFuture;
 
   // constructor necessary to init load / unload state stream
   LifecycleModule() {
+    // The didUnload event must be emitted after disposal which requires that
+    // the stream controller must be disposed of manually at the end of the
+    // unload transition.
+    _didUnloadController = new StreamController<LifecycleModule>.broadcast();
+
     [
       _willLoadController = new StreamController<LifecycleModule>.broadcast(),
       _didLoadController = new StreamController<LifecycleModule>.broadcast(),
       _willUnloadController = new StreamController<LifecycleModule>.broadcast(),
-      _didUnloadController = new StreamController<LifecycleModule>.broadcast(),
       _willLoadChildModuleController =
           new StreamController<LifecycleModule>.broadcast(),
       _didLoadChildModuleController =
@@ -155,6 +159,9 @@ abstract class LifecycleModule extends Object with Disposable {
   final Map<LifecycleModule, StreamSubscription<LifecycleModule>>
       _didUnloadChildModuleSubscriptions = {};
 
+  /// Whether the module is currently instantiated.
+  bool get isInstantiated => _state == LifecycleState.instantiated;
+
   /// Whether the module is currently loaded.
   bool get isLoaded => _state == LifecycleState.loaded;
 
@@ -176,9 +183,6 @@ abstract class LifecycleModule extends Object with Disposable {
   /// Whether the module is currently unloading.
   bool get isUnloading => _state == LifecycleState.unloading;
 
-  /// The current lifecycle state of the module.
-  LifecycleState get state => _state;
-
   //--------------------------------------------------------
   // Public methods that can be used directly to trigger
   // module lifecycle / check current lifecycle state
@@ -189,27 +193,24 @@ abstract class LifecycleModule extends Object with Disposable {
   /// Calls the onLoad() method, which can be implemented on a Module.
   /// Executes the willLoad and didLoad event streams.
   ///
-  /// Initiates the loading process when the module is in the initialized state.
-  /// If the module is in the loaded or loading state a warning is logged and
-  /// the method is a noop. If the module is in any other state, a StateError is
-  /// thrown.
+  /// Initiates the loading process when the module is in the instantiated
+  /// state. If the module is in the loaded or loading state a warning is logged
+  /// and the method is a noop. If the module is in any other state, a
+  /// StateError is thrown.
   ///
   /// Note that [LifecycleModule] only supports one load/unload cycle. If [load]
   /// is called after a module has been unloaded, a [StateError] is thrown.
   Future<Null> load() {
-    if (isLoading) {
-      _logger.warning('Module "$name" is already loading, cannot load.');
-      return _transitionFuture;
+    if (isLoading || isLoaded) {
+      return _buildNoopResponse(
+          isTransitioning: isLoading,
+          methodName: 'load',
+          currentState: LifecycleState.loaded);
     }
 
-    if (isLoaded) {
-      _logger.warning('Module "$name" is already loaded, cannot load.');
-      return new Future.value(null);
-    }
-
-    if (_state != LifecycleState.initialized) {
-      return new Future.error(
-          new StateError('Module "$name" state is $_state, cannot load.'));
+    if (!isInstantiated) {
+      return _buildIllegalTransitionResponse(
+          reason: 'A module can only be loaded once.');
     }
 
     _state = LifecycleState.loading;
@@ -218,8 +219,8 @@ abstract class LifecycleModule extends Object with Disposable {
     _willLoadController.add(this);
 
     onLoad().then((_) {
-      _didLoadController.add(this);
       _state = LifecycleState.loaded;
+      _didLoadController.add(this);
       _transitionFuture = null;
       completer.complete();
     });
@@ -272,19 +273,17 @@ abstract class LifecycleModule extends Object with Disposable {
   /// the method is a noop. If the module is in any other state, a StateError is
   /// thrown.
   Future<Null> suspend() {
-    if (isSuspending) {
-      _logger.warning('Module "$name" is already suspending, cannot suspend.');
-      return _transitionFuture;
-    }
-
-    if (isSuspended) {
-      _logger.warning('Module "$name" is already suspended, cannot suspend.');
-      return new Future.value(null);
+    if (isSuspended || isSuspending) {
+      return _buildNoopResponse(
+          isTransitioning: isSuspending,
+          methodName: 'suspend',
+          currentState: LifecycleState.suspended);
     }
 
     if (!isLoaded) {
-      return new Future.error(
-          new StateError('Module "$name" state is $_state, cannot suspend.'));
+      return _buildIllegalTransitionResponse(
+          targetState: LifecycleState.suspended,
+          allowedStates: [LifecycleState.loaded]);
     }
 
     _state = LifecycleState.suspending;
@@ -294,8 +293,8 @@ abstract class LifecycleModule extends Object with Disposable {
 
     Future.wait(_childModules.map((c) => c.suspend())).then((_) {
       onSuspend().then((_) {
-        _didSuspendController.add(this);
         _state = LifecycleState.suspended;
+        _didSuspendController.add(this);
         _transitionFuture = null;
         completer.complete();
       });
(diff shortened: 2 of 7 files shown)


aviary-wf:
## Raven
Number of Findings: 0

lib/src/lifecycle_module.dart
```diff
@@ -130,12 +156,28 @@ abstract class LifecycleModule extends Object with Disposable {
...
+  bool get isSuspending => _state == LifecycleState.suspending;
+
+  /// Whether the module is currently unloaded.
+  bool get isUnloaded => _state == LifecycleState.unloaded;
+
+  /// Whether the module is currently unloading.
+  bool get isUnloading => _state == LifecycleState.unloading;
+
+  /// The current lifecycle state of the module.
+  LifecycleState get state => _state;
```
maxwellpeterson-wf: Do we need to expose this if we have the getters (above) for each state?

lib/src/lifecycle_module.dart
```diff
@@ -147,24 +189,41 @@ abstract class LifecycleModule extends Object with Disposable {
...
-    } else if (!_isLoaded) {
-      _willLoadController.add(this);
-      onLoad().then((_) {
-        _didLoadController.add(this);
-        _isLoaded = true;
-        completer.complete();
-      });
-    } else {
+    if (isLoading) {
+      _logger.warning('Module "$name" is already loading, cannot load.');
```
maxwellpeterson-wf: Wondering if we can come up with something more helpful than "cannot load".
Maybe something along the lines of:

`.load() was called while Module "$name" is already loading; this is a no-op. Check for any unnecessary calls to .load().`

lib/src/lifecycle_module.dart
```diff
@@ -147,24 +189,41 @@ abstract class LifecycleModule extends Object with Disposable {
...
+      return _transitionFuture;
+    }
+
+    if (isLoaded) {
       _logger.warning('Module "$name" is already loaded, cannot load.');
-      completer.complete();
+      return new Future.value(null);
+    }
```
maxwellpeterson-wf: Would it be worth merging this block a bit with the `isLoading` block to DRY up the logging message?

```dart
if (isLoading || isLoaded) {
  _logger.warning(...);
  if (isLoading) return _transitionFuture;
  else return new Future.value(null);
}
```

I don't have a strong opinion on this so if you'd prefer to keep it how it is for clarity, 👍

lib/src/lifecycle_module.dart
```diff
@@ -207,45 +266,78 @@ abstract class LifecycleModule extends Object with Disposable {
...
+  Future<Null> suspend() {
+    if (isSuspending) {
+      _logger.warning('Module "$name" is already suspending, cannot suspend.');
+      return _transitionFuture;
+    }
+
+    if (isSuspended) {
+      _logger.warning('Module "$name" is already suspended, cannot suspend.');
+      return new Future.value(null);
     }
```
maxwellpeterson-wf: Same question here about DRYing up the logging message (I'll omit this comment from now on, probably applies to all lifecycle methods).

lib/src/lifecycle_module.dart
```diff
@@ -207,45 +266,78 @@ abstract class LifecycleModule extends Object with Disposable {
...
-          'Module "$name" is ${_isLoaded ? 'already suspended' : 'not loaded'}, cannot suspend.');
-      completer.complete();
+  ///
+  /// Initiates the suspend process when the module is in the loaded state. If
+  /// the module is in the suspended or suspending state a warning is logged and
+  /// the method is a noop. If the module is in any other state, a StateError is
+  /// thrown.
+  Future<Null> suspend() {
+    if (isSuspending) {
+      _logger.warning('Module "$name" is already suspending, cannot suspend.');
```
maxwellpeterson-wf: Consider similar alternative to `cannot suspend` depending on how the `cannot load` comment is addressed.

lib/src/lifecycle_module.dart
maxwellpeterson-wf: Note: If it is decided that this getter should be removed, the `LifecycleState` enum no longer needs to be publicly exposed.

maxwellpeterson-wf:
The diagram in the PR description could probably be added to the README

lib/src/lifecycle_module.dart
```diff
@@ -147,24 +189,41 @@ abstract class LifecycleModule extends Object with Disposable {
...
+
+    if (isLoaded) {
       _logger.warning('Module "$name" is already loaded, cannot load.');
-      completer.complete();
+      return new Future.value(null);
+    }
+
+    if (_state != LifecycleState.initialized) {
+      return new Future.error(
+          new StateError('Module "$name" state is $_state, cannot load.'));
```
evanweible-wf: Would be nice to include the reason that the load cannot happen here (i.e. "A module can only be loaded once.")

lib/src/lifecycle_module.dart
```diff
@@ -147,24 +189,41 @@ abstract class LifecycleModule extends Object with Disposable {
...
     }
+
+    _state = LifecycleState.loading;
+    final completer = new Completer<Null>();
+    _transitionFuture = completer.future;
+    _willLoadController.add(this);
+
+    onLoad().then((_) {
+      _didLoadController.add(this);
+      _state = LifecycleState.loaded;
```
evanweible-wf: In practice, I'm not sure this would ever make a difference since the stream is async, but it feels like the `_state` should be updated prior to emitting the event.

lib/src/lifecycle_module.dart
```diff
@@ -207,45 +266,78 @@ abstract class LifecycleModule extends Object with Disposable {
...
-          completer.complete();
-        });
-      });
-    } else {
-      _logger.warning(
-          'Module "$name" is ${_isLoaded ? 'not suspended' : 'not loaded'}, cannot suspend.');
-      completer.complete();
+  ///
+  /// Only initiates the resume process when the module is in the suspended
+  /// state. If the module is in the resuming state a warning is logged and the
```
evanweible-wf: This is slightly different than the semantics for loading. If `load()` is called when the module is already `loaded`, it is a no-op. Here, if the module has been resumed (or was never suspended) and is `loaded`, calling `resume()` will throw. Is that what we want?

lib/src/lifecycle_module.dart
```diff
@@ -274,36 +366,54 @@ abstract class LifecycleModule extends Object with Disposable {
...
+    _transitionFuture = completer.future;
+    _willUnloadController.add(this);
+
+    final unloadChildren = _childModules.map((c) => c.unload());
+    Future.wait(unloadChildren).then((_) {
+      _childModules.clear();
+      onUnload().then((_) {
+        _didUnloadController.add(this);
+        super.dispose().then((_) {
+          _state = LifecycleState.unloaded;
```
evanweible-wf: This seems odd to emit the `didUnload` event and change the `_state` at different times. If disposal is a part of unloading, I think the `didUnload` event should be emitted when that has finished

evanweible-wf: Hmm, okay so there are separate events for disposal, so maybe that is fine, but it still seems odd to have the disconnect between the `didUnload` event and the `unloaded` state.

test/lifecycle_module_test.dart
```diff
@@ -174,218 +222,397 @@ void main() {
...
-      expect(module.isSuspended, isTrue);
-      await module.resume();
-      expect(module.isSuspended, isFalse);
-    });
+    void testInvalidTransitions(
+        LifecycleState state, List<LifecycleState> invalidStates) {
+      invalidStates.forEach((fromState) {
+        test('should throw StateError when state is $fromState', () async {
+          await gotoState(module, fromState);
+          expect(module.state, fromState);
```
evanweible-wf: #nit Most of the `expect()` calls in here use the `equals()` matcher, but this does not. I'm fine with either, but we should be consistent

lib/src/lifecycle_module.dart
evanweible-wf: Similar comment for other invalid transitions below

lib/src/lifecycle_module.dart
bencampbell-wf: The current need to expose this field is to simplify tests. It is easy to write a utility function to relive the need of this property.

I would argue that there is no harm in exposing both the `state` property as well as the `is<LifecycleState>` utility functions as long as the latter is always written in terms of the former.

lib/src/lifecycle_module.dart
bencampbell-wf: @todbachman-wf and I talked about the legality of a state transition and reasoned that a `StateError` is correct whenever a consumer attempts to change states without following a green line in the PR diagram. I guess it is assumed that there is an implicit self referential transition for every state. Your correct in pointing out this is the only update that does not follow this pattern.

I would argue that this should be updated to **NOT** throw **AND** noop  when already in a loaded state.

lib/src/lifecycle_module.dart
bencampbell-wf: I agree. The _didUnload event should be called with respect to a change in state.

lib/src/lifecycle_module.dart
evanweible-wf: Personally I lean towards not including this `state` property because we don't have a use case for it (outside of tests) yet and it forces us to commit to `LifecycleState` as a part of the public API. If we don't do that, then we're only committing to the `bool is[State]` getters.

It's just a lean, though. If others think this adds enough value, I'm cool with it.

bencampbell-wf: Fair enough :) I'm fine either way. I will remove it and update the test utilities.

lib/src/lifecycle_module.dart
```diff
@@ -17,25 +17,55 @@ library w_module.src.lifecycle_module;
...
 
 import 'package:logging/logging.dart';
-import 'package:meta/meta.dart' show protected;
+import 'package:meta/meta.dart' show protected, required;
 import 'package:w_common/disposable.dart';
 
+/// Possible states a [LifecycleModule] may occupy.
+enum LifecycleState {
+  /// The module has been instantiated.
+  initialized,
```
georgelesica-wf: Maybe just change this to `instantiated` to make sure it isn't confused with whatever happens when you call an `init()` method on various modules.

lib/src/lifecycle_module.dart
```diff
@@ -17,25 +17,55 @@ library w_module.src.lifecycle_module;
...
   Logger _logger;
   String _name = 'Module';
+  LifecycleState _state = LifecycleState.initialized;
+  Future<Null> _transitionFuture;
 
   // constructor necessary to init load / unload state stream
   LifecycleModule() {
+    // The didUnload event must be emitted after disposal which requires that
+    // the stream controller must be disposed of manually at the end of the
+    // unload transition.
```
georgelesica-wf: Tricky!

lib/src/lifecycle_module.dart
```diff
@@ -147,24 +193,38 @@ abstract class LifecycleModule extends Object with Disposable {
   /// Calls the onLoad() method, which can be implemented on a Module.
   /// Executes the willLoad and didLoad event streams.
   ///
-  /// [LifecycleModule] only supports one load/unload cycle. If [load] is called
-  /// again after a module has been unloaded, a [StateError] is thrown.
+  /// Initiates the loading process when the module is in the initialized state.
```
georgelesica-wf: If we rename, don't miss this one.

lib/src/lifecycle_module.dart
```diff
@@ -207,45 +267,80 @@ abstract class LifecycleModule extends Object with Disposable {
...
+
+    _state = LifecycleState.suspending;
+    final completer = new Completer<Null>();
+    _transitionFuture = completer.future;
+    _willSuspendController.add(this);
+
+    Future.wait(_childModules.map((c) => c.suspend())).then((_) {
+      onSuspend().then((_) {
+        _didSuspendController.add(this);
+        _state = LifecycleState.suspended;
```
georgelesica-wf: #nit Can we set the state first? Ultimately it won't matter but just to keep things consistent and clearly express what we're trying to do it makes sense in my mind.

lib/src/lifecycle_module.dart
georgelesica-wf: It's context-sensitive, right? Like, if you're in `loaded`, you know you got there through `loading` because there's no other way. But when you're in `loaded` you haven't necessarily ever visited `resuming`. So the two transitions are slightly different in that sense.

lib/src/lifecycle_module.dart
```diff
@@ -274,36 +369,54 @@ abstract class LifecycleModule extends Object with Disposable {
...
+          currentState: LifecycleState.unloaded);
+    }
+
+    if (!(isLoaded || isSuspended)) {
+      return _buildIllegalTransitionResponse(
+          targetState: LifecycleState.unloaded,
+          allowedStates: [LifecycleState.loaded, LifecycleState.suspended]);
+    }
+
+    ShouldUnloadResult canUnload = shouldUnload();
```
georgelesica-wf: Can we call this something like `shouldUnloadResult` or something? `canUnload.shouldUnload` is bugging me...

georgelesica-wf:
+1

test/lifecycle_module_test.dart
```diff
@@ -174,218 +254,468 @@ void main() {
...
+      test('should return a useful future if it is already loading', () async {
+        await gotoState(module, LifecycleState.loading);
+        expect(module.isLoading, isTrue);
+        var future = module.load();
+        expect(module.isLoaded, isFalse);
+        await future;
+        expect(module.isLoaded, isTrue);
+      });
+
+      test('should return pending future if it is already loading', () async {
```
evanweible-wf: This test and the one above are essentially duplicates. Fwiw, I like this one better,

test/lifecycle_module_test.dart
```diff
@@ -174,218 +254,468 @@ void main() {
...
+
+      test('should warn if it is already loading', () async {
+        await gotoState(module, LifecycleState.loading);
+        expect(lastLogMessage, isNull);
+        await module.load();
+        expect(lastLogMessage, isNotNull);
+        expect(lastLogMessage.level, equals(Level.WARNING));
+      });
+
+      test('should do nothing if it was already loaded', () async {
```
evanweible-wf: I don't think this test covers anything that the "should only load once if it is already loading" test doesn't.

test/lifecycle_module_test.dart
evanweible-wf: Same thing for similar test cases below

test/lifecycle_module_test.dart
evanweible-wf: similar scenarios below

README.md
```diff
@@ -295,6 +295,15 @@ Method         | Description
 `shouldUnload` | Returns the unloadable state of the `Module` and its child modules as a `ShouldUnloadResult`.  Internally, this executes the module's `onShouldUnload` method.
 `unload`       | Triggers unloading of a `Module` and all of its child modules.  Internally, this executes the module's `shouldUnload` method, and, if that completes successfully, executes the module's `onUnload` method. If unloading is rejected, this method will complete with an error.
 
+![lifecycle module - lifecycle](https://cloud.githubusercontent.com/assets/11619752/21526229/d7a5ae1c-ccdf-11e6-9e95-ff16a83e3a7f.png)
```
bencampbell-wf: This is the method that the other image in the README are using. Given the content delivery mechanism is immutable, the HTTP references to them are effectively versioned.

evanweible-wf:
+1

evanweible-wf:
+10 
- [x] CI passes, tests updated and reviewed

georgelesica-wf:
+1

evanweible-wf:
+1

georgelesica-wf:
+1

evanweible-wf:
+10 
- [x] CI passes, tests updated and reviewed

evanweible-wf:
QA +1
- [x] Testing instruction
- [x] Dev +1's
- [x] All unit tests pass

Merging into master.

PR #65, opened 2016-12-28
RAP-1501 Implement Disposal Via Composition

bencampbell-wf:
Problem
---

Disposal of a `LifecycleModule` has been encapsulated as part of the `unload` lifecycle call. In order to provide the disposable management functions to child classes `LifecycleModule` was force to mixin `Disposable` and add deprecated aliases to the 'dispose' and 'onDispose' methods to 'unload' and 'onUnload' respectively. This extends the API of `LifecycleModule` in an inappropriate way that may confuse consumers. In addition, a `LifecycleModule` may be inadvertently maintained in a collection of `Disposable`s or managed by another `Disposable`.

Solution
---

Utilize the `DisposableManager` interface to define the disposable management API for a `LifecycleModule` without having to implement 'dispose' and `onDispose`. Calls to these management functions are proxied to an internal `Disposable` object which the `LifecycleModule` disposes as part of it's `unload` lifecycle call. This ensures child classes can continue to manage disposable entities and that their disposal is done during the unload lifecycle transition as defined by the `LifecycleModule`.

Areas of Impact
---

- Removes deprecated methods from API
- Disposal

Howto QA/+10
---

- [ ] Ensure CI passes

References
---

- /fya @Workiva/web-platform-pp @Workiva/rich-app-platform-pp

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -49,7 +49,8 @@ enum LifecycleState {
 
 /// Intended to be extended by most base module classes in order to provide a
 /// unified lifecycle API.
-abstract class LifecycleModule extends Object with Disposable {
+abstract class LifecycleModule implements DisposableManager {
+  final Disposable _disposableProxy = new Disposable();
   Logger _logger;
   String _name = 'Module';
   LifecycleState _state = LifecycleState.instantiated;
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Raven
Number of Findings: 0

pubspec.yaml
```diff
@@ -27,5 +27,10 @@ dev_dependencies:
   test: ^0.12.0
   w_flux: ^1.0.0
 
+dependency_overrides:
+  w_common:
+    git:
+      url: https://github.com/Workiva/w_common.git
+      ref: 33da10e3790f9fb685122393f4321422e5b178c5
```
bencampbell-wf: Note: This is a WIP until https://github.com/Workiva/w_common/pull/17 is merged and released.

lib/src/lifecycle_module.dart
```diff
@@ -49,7 +49,8 @@ enum LifecycleState {
 
 /// Intended to be extended by most base module classes in order to provide a
 /// unified lifecycle API.
-abstract class LifecycleModule extends Object with Disposable {
+abstract class LifecycleModule implements DisposableManager {
+  final Disposable _disposable = new Disposable();
```
georgelesica-wf: Maybe `_disposableProxy` or something? I dunno, if other people are fine with the name I'm down.

lib/src/lifecycle_module.dart
```diff
@@ -262,6 +263,32 @@ abstract class LifecycleModule extends Object with Disposable {
     return completer.future;
   }
 
+  /// Ensures a given [Disposable] is disposed when the [LifecycleModule] is
+  /// unloaded.
+  @override
+  @protected
```
georgelesica-wf: I don't think we should actually make these `@protected` any more since then it technically doesn't implement `DisposableManager`, right? It's weird. I don't really like computers.

georgelesica-wf:
+1

trentgrover-wf:
+1
glad this snuck in before a release with the disposable mixing :)

maxwellpeterson-wf:
+1

todbachman-wf:
+10

This is a really nice improvement over my previous hack!

todbachman-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [ ] Security review (if needed)
  - N/A
- [x] All unit tests pass
- [x] Rosie reports a clean dependency scan

Merging into master.

PR #68, opened 2017-02-03
RAP-1529 Do not allow loadChildModule calls when unloading or unloaded

bencampbell-wf:
Problem
---

If a module attempts to load a child module while unloading or unloading it may be possible for internal event stream controllers to be added to after they have been closed. This returns an unxepcted exception which may be confusing for consumers.

Solution
---

By making calls to loadChildModule when a module is unloaded or unloading throw a state error indicating it is not allowed consumers better understand the failure.

Howto QA/+10
---

- [ ] Ensure CI passes

References
---

- /fya @Workiva/rich-app-platform-pp @Workiva/web-platform-pp

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -232,16 +232,19 @@ abstract class LifecycleModule implements DisposableManager {
 
   /// Public method to async load a child module and register it
   /// for lifecycle management.
+  ///
+  /// Attempting to load a child module after a module has been unloaded will
+  /// throw a [StateError].
   @protected
   Future<Null> loadChildModule(LifecycleModule newModule) {
     if (_childModules.contains(newModule)) {
-      return new Future(() {});
+      return new Future.value(null);
     }
 
     if (isUnloaded || isUnloading) {
-      final state = isUnloaded ? 'unloaded' : 'unloading';
-      return new Future.error(
-          new StateError('Cannot load child module when module is $state'));
+      var stateLabel = isUnloaded ? 'unloaded' : 'unloading';
+      return new Future.error(new StateError(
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

lib/src/lifecycle_module.dart
```diff
@@ -237,6 +237,13 @@ abstract class LifecycleModule implements DisposableManager {
     if (_childModules.contains(newModule)) {
       return new Future(() {});
```
georgelesica-wf: You didn't do this (I may have), but a better way is `Future.value(null)`, could we go ahead and change it while you're fiddling in this file?

lib/src/lifecycle_module.dart
```diff
@@ -237,6 +237,13 @@ abstract class LifecycleModule implements DisposableManager {
...
     }
+
+    if (isUnloaded || isUnloading) {
+      final state = isUnloaded ? 'unloaded' : 'unloading';
```
georgelesica-wf: There was some discussion in the Dart chat about defaulting to `var` again unless there is a particular reason to use `final` now that the analyzer can detect type-unstable reassignments.

georgelesica-wf:
+1 with just a couple questions.

evanweible-wf:
+1

todbachman-wf:
+10. Confirmed CI passed.

todbachman-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [ ] Security review (if needed)
  - N/A
- [x] All unit tests pass
- [x] Rosie reports a clean dependency scan

Merging into master.

PR #81, opened 2017-04-19
RAP-1850 Await Loading, Suspending, and Resuming State Transitions

bencampbell-wf:
## Problem

The current implementation does not allow state transitions to occur while a transition is in progress. For instance, a consumer cannot call suspend and unload synchronously as unloading is not allowed when a module is suspending. Unloading **is** valid when a module is suspended though. If a consumer fails to await the suspend, the LifecycleModule should wait for the transition to complete and then proceed to unload.

## Solution

Any transitioning state that is resolving to a valid state should be awaited in the event of a state change. For example, if a consumer `suspend`s and `unload`s in the same synchronous block, the unload should await the completion of the suspension before proceeding. The state of the module changes from isSuspending  to isUnloading so subsequent state changes must respect legal transitions. The lifecycle events expected from suspend and unload fire in order as if the consumer had awaited the suspend call before unloading.

transitions

The implicit transitions illustrated above are now allowed as permitted by their associated explicit transition. Unload will wait for loading, suspending, and resuming transitions to complete before unloading. Suspend and resume will await resuming and suspending transitions respectively.

## Howto Test
- [ ] CI Passes

## References
- /fya @Workiva/rich-app-platform-pp @Workiva/web-platform-pp

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -277,18 +277,8 @@ abstract class LifecycleModule extends SimpleModule
 
     _state = LifecycleState.loading;
     _transition = new Completer<Null>();
-    _willLoadController.add(this);
 
-    onLoad().then((_) {
-      _state = LifecycleState.loaded;
-      _didLoadController.add(this);
-      _transition.complete();
-      _transition = null;
-    }).catchError((Object error, StackTrace stackTrace) {
-      _didLoadController.addError(error, stackTrace);
-      _transition.completeError(error, stackTrace);
-      _transition = null;
-    });
+    _load().then(_transition.complete).catchError(_transition.completeError);
 
     return _transition.future;
   }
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Raven
Number of Findings: 0

maxwellpeterson-wf:
+1 lgtm

lib/src/lifecycle_module.dart
```diff
@@ -415,30 +405,19 @@ abstract class LifecycleModule extends SimpleModule
...
-    Future.wait(suspendingChildren).then((_) async {
-      await onSuspend();
-      _state = LifecycleState.suspended;
-      _didSuspendController.add(this);
-      _transitionFuture = null;
-      completer.complete();
-    }).catchError((Object error, StackTrace stackTrace) {
-      _didSuspendController.addError(error, stackTrace);
-      completer.completeError(error, stackTrace);
-    });
```
bencampbell-wf: This functionality has been [defined](https://github.com/Workiva/w_module/pull/81/files#diff-c586827f6717f60db651d466b5ab2b5fR637) as pure async private helper method.

maxwellpeterson-wf:
+1

georgelesica-wf:
+1

maxwellpeterson-wf:
+10
- CI passes and new tests cover src changes.

maxwellpeterson-wf:
@jayudey-wf ready for you

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - performed and documented by Max
- [ ] All unit tests pass
- [ ] **Dependency Scan Clean**

Merging into master.

PR #86, opened 2017-06-13
RAP-2057 Fix double unload bug

georgelesica-wf:
[Jira Ticket](https://jira.atl.workiva.net/browse/RAP-2057)

## Problem

A bug was reported by @josephelliot-wk in which LifecycleModule throws if an unload was attempted and rejected because a child module returned false from its shouldUnload method.
The problem is that we are leaving the module in the unloading state after a rejected unload when we should probably be moving it back into the loaded state (or not moving it to unloading until after the shouldUnload check has passed).

## Solution

Ensure that the state is reset to its previous state if the unload is rejected. We have to do this in the async `_unload` to keep the semantics the same as they were previously (transition to unloading happens immediately).

I updated one of the tests to guard against regressions.

## Potential Regressions

State changes. These are well tested.

## Tests

Added and/or updated.

## QA / +10

1. Checkout the branch.
2. Update dependencies.
3. Unit tests should pass.
4. Solves the problem @josephelliot-wk identified.

## FYI

@josephelliot-wk

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -758,8 +758,7 @@ abstract class LifecycleModule extends SimpleModule
       }
       _didUnloadController.add(this);
       await _didUnloadController.close();
-    } on ModuleUnloadCanceledException catch (error, stackTrace) {
-      _didUnloadController.addError(error, stackTrace);
+    } on ModuleUnloadCanceledException catch (error, _) {
       rethrow;
     } catch (error, stackTrace) {
       _didUnloadController.addError(error, stackTrace);
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -687,9 +687,11 @@ abstract class LifecycleModule extends SimpleModule
         await pendingTransition;
       }
       _willResumeController.add(this);
+      List<Future<Null>> childResumeFutures = <Future<Null>>[];
       for (var child in _childModules.toList()) {
-        await child.resume();
+        childResumeFutures.add(child.resume());
       }
+      await Future.wait(childResumeFutures);
       await onResume();
       if (_state == LifecycleState.resuming) {
         _state = LifecycleState.loaded;
@@ -708,9 +710,11 @@ abstract class LifecycleModule extends SimpleModule
         await pendingTransition;
       }
       _willSuspendController.add(this);
+      List<Future<Null>> childSuspendFutures = <Future<Null>>[];
       for (var child in _childModules.toList()) {
-        await child.suspend();
+        childSuspendFutures.add(child.suspend());
       }
+      await Future.wait(childSuspendFutures);
       await onSuspend();
       if (_state == LifecycleState.suspending) {
         _state = LifecycleState.suspended;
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Raven
Number of Findings: 0

lib/src/lifecycle_module.dart
```diff
@@ -725,25 +731,30 @@ abstract class LifecycleModule extends SimpleModule
...
+      _childModules.clear();
       await onUnload();
       await _disposableProxy.dispose();
       if (_state == LifecycleState.unloading) {
         _state = LifecycleState.unloaded;
+        _previousState = null;
         _transition = null;
       }
       _didUnloadController.add(this);
+      await _didUnloadController.close();
```
bencampbell-wf: Removing this from the try block introduces a possible leak in the event of an exception. The behavior of the module may be undefined in this case, but we should be sure to cleanup the stream controller to prevent leaks.

lib/src/lifecycle_module.dart
```diff
@@ -725,25 +735,36 @@ abstract class LifecycleModule extends SimpleModule
...
       if (_state == LifecycleState.unloading) {
         _state = LifecycleState.unloaded;
+        _previousState = null;
         _transition = null;
       }
       _didUnloadController.add(this);
-    } catch (error, stackTrace) {
+      await _didUnloadController.close();
+    } on ModuleUnloadCanceledException catch (error, stackTrace) {
       _didUnloadController.addError(error, stackTrace);
```
bencampbell-wf: I would suggest not adding `ModuleUnloadCanceledException` errors to the didUnload event stream. This is semantically correct seeing as the module will revert to the previous state if the ShouldUnloadResult returns false. 

Emitting this event would be a lie as the module has *not* unloaded.

georgelesica-wf: Agreed. My only concern is that someone is relying on this behavior since, I think, it predates this PR. I'm cool with calling it a bug, though, which (I think) allows us to change it with a patch or minor release.

georgelesica-wf:
@Workiva/performance-reviewers

bencampbell-wf:
+10

Things Tested
-------------
- [x] Checkout the branch.
- [x] Update dependencies.
- [x] Unit tests should pass.
- [x] Solves the problem @josephelliot-wk identified.

bencampbell-wf:
@jayudey-wf ready for QA.

bencampbell-wf:
Lets hold of one moment, @jayudey-wf .

@georgelesica-wf is going to address the remaining comment.

lib/src/lifecycle_module.dart
georgelesica-wf: Update: I've decided to change this behavior and consider it a bug. The key piece fact is that when a `shouldUnload` rejects an unload, the `willUnload` is not dispatched, meaning that if we add the error to the `didUnload` we're actually breaking the lifecycle itself. When the `shouldUnload` rejects, no streams will dispatch, but the error will be allowed to propagate to the `unload` call.

bencampbell-wf:
+10

Things Tested
-------------
- [x] Checkout the branch.
- [x] Update dependencies.
- [x] Unit tests should pass.
- [x] Solves the problem @josephelliot-wk identified.

bencampbell-wf:
@jayudey-wf ready for QA.

bencampbell-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [x] All unit tests pass

Merging into master.

PR #90, opened 2017-07-07
RAP-2055 Log LifecycleEvents

bencampbell-wf:
## Problem

There is no insight into the current lifecycle state for a instance of LifecycleModule. 

## Solution

Emit a log event when a `LifecycleModule` transitions to a new lifecycle state. This provides a consistent and clear indication of what state any given LifecycleModule instance is in.

## Howto QA

- [ ] Ensure CI passes
    - Explicit unit tests were added to confirm behavior
- [ ] Link into an existing project
- [ ] Set the log level to at least FINE
- [ ] Confirm lifecycle state changes are logged

## References
/fya @Workiva/rich-app-platform-pp @Workiva/web-platform-pp

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -64,7 +64,7 @@ abstract class LifecycleModule extends SimpleModule
   StreamController<LifecycleModule> _didUnloadController;
   final Disposable _disposableProxy = new Disposable();
   Logger _logger;
-  String _name = 'Module';
+  String _name;
   LifecycleState _previousState;
   LifecycleState _state = LifecycleState.instantiated;
   Completer<Null> _transition;
@@ -103,12 +103,28 @@ abstract class LifecycleModule extends SimpleModule
       _didResumeController = new StreamController<LifecycleModule>.broadcast()
     ].forEach(manageStreamController);
 
-    _logger = new Logger('w_module');
+    _logger = new Logger('$name');
+    <
+        String,
+        Stream>{
+      'didLoad': didLoad,
+      'didLoadChildModule': didLoadChildModule,
+      'didResume': didResume,
+      'didSuspend': didSuspend,
+      'didUnload': didUnload,
+      'didUnloadChildModule': didUnloadChildModule,
+      'willLoad': willLoad,
+      'willLoadChildModule': willLoadChildModule,
+      'willResume': willResume,
+      'willSuspend': willSuspend,
+      'willUnload': willUnload,
+      'willUnloadChildModule': willUnloadChildModule,
+    }.forEach(_logLifecycleEvents);
   }
 
   /// Name of the module for identification in exceptions and debug messages.
   // ignore: unnecessary_getters_setters
-  String get name => _name;
+  String get name => _name ?? '$runtimeType';
 
   /// Deprecated: the module name should be defined by overriding the getter in
   /// a subclass and it should not be mutable.
@@ -659,6 +675,14 @@ abstract class LifecycleModule extends SimpleModule
       _didLoadController.addError(error, stackTrace);
       rethrow;
     }
+  }
+
+  /// A utility to logging LifecycleModule lifecycle events
+  void _logLifecycleEvents(
+      String logLabel, Stream<dynamic> lifecycleEventStream) {
+    manageStreamSubscription(lifecycleEventStream.listen(
+        (_) => _logger.fine(logLabel),
+        onError: (error) => _logger.warning('$logLabel error: $error')));
   }
 
   /// Handles a child [LifecycleModule]'s [didUnload] event.
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -63,6 +63,7 @@ abstract class LifecycleModule extends SimpleModule
       _didUnloadChildModuleSubscriptions = {};
   StreamController<LifecycleModule> _didUnloadController;
   final Disposable _disposableProxy = new Disposable();
+  final Disposable _postUnloadDisposable = new Disposable();
   Logger _logger;
   String _name;
   LifecycleState _previousState;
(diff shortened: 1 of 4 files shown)


aviary2-wf:
## Raven
<details>

</details>

lib/src/lifecycle_module.dart
```diff
@@ -103,12 +103,28 @@ abstract class LifecycleModule extends SimpleModule
       _didResumeController = new StreamController<LifecycleModule>.broadcast()
     ].forEach(manageStreamController);
 
-    _logger = new Logger('w_module');
+    _logger = new Logger('$name');
+    <
```
bencampbell-wf: This is what the formatter insists.

georgelesica-wf: Nobody let Greg or Aaron see this PR!

lib/src/lifecycle_module.dart
```diff
@@ -103,12 +103,28 @@ abstract class LifecycleModule extends SimpleModule
...
+      'willSuspend': willSuspend,
+      'willUnload': willUnload,
+      'willUnloadChildModule': willUnloadChildModule,
+    }.forEach(_logLifecycleEvents);
   }
 
   /// Name of the module for identification in exceptions and debug messages.
   // ignore: unnecessary_getters_setters
-  String get name => _name;
+  String get name => _name ?? '$runtimeType';
```
georgelesica-wf: I've been told that `runtimeType` gets hosed in JS, but I've never actually seen it not work, even in JS (probably only problematic in prod builds). What does it end up as in those cases? Is this still a reasonable default in that case?

test/lifecycle_module_test.dart
```diff
@@ -283,9 +283,12 @@ Future<Null> executeStateTransition(
 
 void main() {
   Logger.root.level = Level.ALL;
+  var messagePattern = new RegExp(r'^did|^will');
 
   LogRecord lastLogMessage;
-  Logger.root.onRecord.listen((LogRecord message) {
+  Logger.root.onRecord
+      .where((LogRecord r) => !messagePattern.hasMatch(r.message))
+      .listen((LogRecord message) {
```
georgelesica-wf: Any value in adding tests around the new log messages?

lib/src/lifecycle_module.dart
bencampbell-wf: I hadn't heard that. I will test in a JS browser and adjust accordingly.

test/lifecycle_module_test.dart
bencampbell-wf: It would prevent against accidental reversion.

lib/src/lifecycle_module.dart
bencampbell-wf: You are correct. If minification is applied `runtimeType` will return the minified symbol name.

lib/src/lifecycle_module.dart
```diff
@@ -103,12 +101,33 @@ abstract class LifecycleModule extends SimpleModule
...
+      'willSuspend': willSuspend,
+      'willUnload': willUnload,
+      'willUnloadChildModule': willUnloadChildModule,
+    }.forEach(_logLifecycleEvents);
   }
 
   /// Name of the module for identification in exceptions and debug messages.
   // ignore: unnecessary_getters_setters
-  String get name => _name;
+  String get name => _name ?? 'LifecycleModule($runtimeType)';
```
bencampbell-wf: @georgelesica-wf I updated the default name to use a recognizable string that provides the instance symbol when available.

The release notes should be updated to instruct consumers to override `name` to ensure their log statements always include a recognizable symbol.

test/lifecycle_module_test.dart
```diff
@@ -50,11 +52,11 @@ class TestLifecycleModule extends LifecycleModule {
   List<String> eventList;
   bool mockShouldUnload;
 
-  TestLifecycleModule()
+  TestLifecycleModule({String name})
       : managedDisposable = new Disposable(),
         managedStreamController = new StreamController<Null>(),
         managedStreamSubscription = new MockStreamSubscription(),
-        name = 'TestLifecycleModule' {
+        this.name = name {
```
georgelesica-wf: The setter is actually deprecated, we probably shouldn't use it. Maybe provide a named constructor that accepts a name?

bencampbell-wf: The test class overrides the name getter and provides a setter in the test implementation. This is not using the deprecated setter in the base implementation.

georgelesica-wf: Hmm, yeah, I guess I was wondering about the pattern that we're going to recommend to consumers.

bencampbell-wf: Thats fair.  I have updated the test implementation to remove the setter and use named.

test/lifecycle_module_test.dart
```diff
@@ -32,6 +32,7 @@ class MockStreamSubscription extends Mock implements StreamSubscription<Null> {}
 class TestLifecycleModule extends LifecycleModule {
   Iterable<StreamSubscription<LifecycleModule>> _eventListStreamSubscriptions;
   bool _managedDisposerWasCalled = false;
+  String _name;
```
georgelesica-wf: Oh, see, I was hoping for a named constructor on the `LifecycleModule` class. I don't think consumers should have to override a private attribute to set their own name.

test/lifecycle_module_test.dart
```diff
@@ -50,11 +52,11 @@ class TestLifecycleModule extends LifecycleModule {
   List<String> eventList;
   bool mockShouldUnload;
 
-  TestLifecycleModule()
+  TestLifecycleModule({String name})
       : managedDisposable = new Disposable(),
         managedStreamController = new StreamController<Null>(),
         managedStreamSubscription = new MockStreamSubscription(),
-        name = 'TestLifecycleModule' {
+        name = name ?? 'TestLifecycleModule' {
```
bencampbell-wf: @georgelesica-wf This sets a final String property on which overrides the base implementation for the name getter.

georgelesica-wf:
+1 looks good, thanks for making those changes!

bencampbell-wf:
@jayudey-wf Ready for QA.

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - passing CI w/ tests exercising new functionality
  - installed into another project w/ dependency overrides and verified that all logging events were being emitted as expected
- [ ] All unit tests pass

Merging into master.

PR #91, opened 2017-07-11
RAP-2139 Ensure dispose is called when onUnload fails.

smaifullerton-wk:
### Description

The call to the internal disposable proxy is made after `onUnload`. If an error is returned by `onUnload`, this dispose call is never made leaking all instances retained by `Disposable`.

### Changes

Dispose `_disposableProxy` in the `onUnload` error catch block.

### Semantic Versioning

- [x] **Patch**
  - [x] This change does not affect the public API
  - [ ] This change fixes existing incorrect behavior without any additions
- [ ] **Minor**
  - [ ] This change adds something to the public API
  - [ ] This change deprecates a part of the public API
* [ ] Major
  - [ ] This change modifies part of the public API in a backwards-incompatible manner
  - [ ] This change removes part of the public API

### Testing/QA

- [ ] CI passes

### Code Review

@Workiva/rich-app-platform-pp

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -801,9 +801,12 @@ abstract class LifecycleModule extends SimpleModule
       rethrow;
     } catch (error, stackTrace) {
       _didUnloadController.addError(error, stackTrace);
-      await _disposableProxy.dispose();
-      await _postUnloadDisposable.dispose();
-      rethrow;
+      try {
+        await _disposableProxy.dispose();
+      } finally {
+        await _postUnloadDisposable.dispose();
+        rethrow;
+      }
     }
   }
 }
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -801,6 +801,7 @@ abstract class LifecycleModule extends SimpleModule
       rethrow;
     } catch (error, stackTrace) {
       _didUnloadController.addError(error, stackTrace);
+      await _disposableProxy.dispose();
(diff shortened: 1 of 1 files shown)


aviary2-wf:
## Raven
<details>

</details>

lib/src/lifecycle_module.dart
```diff
@@ -801,6 +801,7 @@ abstract class LifecycleModule extends SimpleModule
       rethrow;
     } catch (error, stackTrace) {
       _didUnloadController.addError(error, stackTrace);
+      await _disposableProxy.dispose();
```
bencampbell-wf: I would be concerned that the call to `_disposableProxy.dispose()` above was the source of the exception that is being caught in this bock. A second attempt to call it would *probably* result in another exception preventing `_postUnlaodDisposable.dispose()` from being called.

georgelesica-wf: Could wrap this call in a nested `try` block and then put lines 805 and 806 in its `finally` clause...?

bencampbell-wf: That would work.

georgelesica-wf:
+10 CI passed

georgelesica-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [ ] Unit tests created/updated - change would be very difficult to test, low-value
- [x] All unit tests pass
- [ ] Rosie has run and reports properly the release the ticket will be included in - NA, OSS
- [x] Security review completed if necessary

Merging into master.

PR #95, opened 2017-08-08
WP-4899 Allow disposal of LifecycleModule prior to load

evanweible-wf:
## Issue

It is possible that a `LifecycleModule` may have some things that need to be managed immediately after construction (i.e. before load). Currently, you cannot dispose/unload a module that has not been loaded, which makes it impossible to clean up these things.

## Solution

Allow `unload()` to proceed if the module is still in the `isInstantiated` state (not loaded) by short-circuiting directly to disposal.

## Testing

- [ ] CI passes (still need to add tests)

## Code Review

@Workiva/web-platform-pp @Workiva/rich-app-platform-pp

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -583,6 +583,13 @@ abstract class LifecycleModule extends SimpleModule
               isUnloading ? LifecycleState.unloading : LifecycleState.unloaded);
     }
 
+    if (isInstantiated) {
+      _state = LifecycleState.unloaded;
+      _previousState = null;
+      _transition = null;
+      return _dispose();
+    }
+
     if (!(isLoaded || isLoading || isResuming || isSuspended || isSuspending)) {
       return _buildIllegalTransitionResponse(
           targetState: LifecycleState.unloaded,
@@ -694,6 +701,14 @@ abstract class LifecycleModule extends SimpleModule
     return _transition?.future ?? new Future.value(null);
   }
 
+  Future<Null> _dispose() async {
+    try {
+      await _disposableProxy.dispose();
+    } finally {
+      await _postUnloadDisposable.dispose();
+    }
+  }
+
   Future<Null> _load() async {
     try {
       _willLoadController.add(this);
@@ -796,6 +811,7 @@ abstract class LifecycleModule extends SimpleModule
   }
 
   Future<Null> _unload(Future<Null> pendingTransition) async {
+    var unloadWasCanceled = false;
     try {
       if (pendingTransition != null) {
         await pendingTransition;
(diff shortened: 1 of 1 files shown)


aviary2-wf:
## Raven
<details>

</details>

georgelesica-wf:
Are there any concrete examples of situations where this would be necessary?

srinivasdhanwada-wf:
@georgelesica-wf https://github.com/Workiva/truss/pull/849 has an example of this, along with a workaround.

evanweible-wf:
@georgelesica-wf this was the motivation for me: https://github.com/Workiva/wdesk_sdk/pull/1588/files#r131959349 (may need to expand the `rich_experience_manager.dart` file, it was collapsing for me)

georgelesica-wf:
I guess I don't quite understand why this stuff can't be done in the `onLoad` callback... https://github.com/Workiva/truss/blob/0948896bc9e519fb56cd8fa7fd04cbf0f9f814c4/lib/src/progress_manager/progress_manager_module.dart

I mean, I don't disagree with this change in principle, it just seems a little unfortunate that we need to do this, so I want to make sure we ponder it sufficiently.

todbachman-wf:
It would probably be worth updating the state diagram in the README to reflect this change.

srinivasdhanwada-wf:
@georgelesica-wf Is that in regards to https://github.com/Workiva/truss/blob/0948896bc9e519fb56cd8fa7fd04cbf0f9f814c4/lib/src/progress_manager/progress_manager_module.dart ?

I don't think there's anything serious that's stopping those lines from being moved to the `onLoad` callback, but there a couple of reasons I like having them in the constructor:

A constructor is responsible for building and returning an instance of the object. In my mind, that includes setting up any variables that exist. If we were to move those first few lines into the `onLoad` callback, we aren't really constructing the instance until the `load` step. What would be the point of the instantiated state? 

The `onLoad` callback is used for setting up event listeners, child modules, stream subscriptions, etc. Anything module needs when "loaded." I could see an argument for moving the `manageDisposable` call, but I think it's clearer to manage these objects as soon as we create them.

I think the module in this case is not a very good use case of a module since the instance itself is not really responsible for anything. The `ProgressManager` creates an instance of the module is really responsible for manipulating the store/events through api calls. Internal instances of the actions, stores, events, and api could be constructed, managed and used within the `ProgressManager` itself without needing the module. Regardless, this example is probably not the only case of the pattern existing and I think there should still be a way to properly manage the disposal of the module, even if it never gets loaded.

georgelesica-wf:
@srinivasdhanwada-wf that sounds reasonable to me. But I'd still like to see actual situations where weirding up the lifecycle will be worth it to get better reliability / clarity somewhere else. If we shouldn't be using a module in this particular case then that seems like a bad reason to change the module lifecycle in order to make this case work better.

trentgrover-wf:
any further plans for this beyond fixing the 1 failing unit test @evanweible-wf ?

evanweible-wf:
I was just making sure that there wasn't any way I could get around needing to dispose a module still in the isInstantiated state in order to appease George, but I never found a good way. I agree that it's less than ideal to further complicate the lifecycle, but I think it makes sense to allow unload/disposal at any point, otherwise it becomes a weird requirement that things usually done in the constructor be moved to `onLoad()` – one that can be incompatible with other requirements (like the experience proxy pattern I landed on in wdesk_sdk).

evanweible-wf:
Closing in favor of https://github.com/Workiva/w_module/pull/101

PR #97, opened 2017-08-09
WP-4919 Unblock transition errors

toddbeckman-wf:
# Problem

If a module has an uncaught exception during an error, it is impossible to unload even after #95 is consumed to permit unloading from instantiated state.

# Solution

# Testing Suggestion

TBA


aviary-wf:
## Raven
<details>

</details>

toddbeckman-wf:
Closing in favor of #100 to prevent dependency on #95

PR #99, opened 2017-08-11
WP-4883 Ensure Examples and Tests run w/ DDC

srinivasdhanwada-wf:
### Description
w_module has not been updated to run with DDC

### Changes
Update w_module to be DDC compliant!
- Updated version constraints on `w_flux` and `react` to pull in DDC compliant verions
- Updated examples to compile with DDC
- Addressed test failures when using DDC
- Addressed deprecations

### Testing/QA
- [ ] CI Passes
- [ ] Examples run w/ DDC
  - [ ] `pub serve example --web-compiler=dartdevc`, then navigate to localhost:8080 in chrome
- [ ] Tests run w/ DDC
  - [ ] `pub run dart_dev test --pub-serve --web-compiler=dartdevc -p chrome -p vm` 

### Review
@Workiva/web-platform-pp

### DIFF
--- a/lib/src/serializable.dart
+++ b/lib/src/serializable.dart
@@ -70,14 +70,6 @@ class _ModuleRegistration extends Object with Disposable {
   final SerializableModule module;
 
   _ModuleRegistration(this.module);
-
-  @override
-  StreamSubscription<T> listenToStream<T>(
-      Stream<T> stream, void onData(T event),
-      {Function onError, void onDone(), bool cancelOnError}) {
-    return super.listenToStream(stream, onData,
-        onError: onError, onDone: onDone, cancelOnError: cancelOnError);
-  }
 }
 
 class SerializableBus {
--- a/lib/src/serializable.dart
+++ b/lib/src/serializable.dart
@@ -36,8 +36,11 @@ abstract class Bridge<T> extends Object with Disposable {
   void broadcastSerializedEvent(Map event);
   void handleSerializedApiCall(T apiCall);
 
-  void _manageStreamSubscription(StreamSubscription subscription) {
-    super.manageStreamSubscription(subscription);
+  StreamSubscription<T> _listenToStream<T>(
+      Stream<T> stream, void onData(T event),
+      {Function onError, void onDone(), bool cancelOnError}) {
+    return super.listenToStream(stream, onData,
+        onError: onError, onDone: onDone, cancelOnError: cancelOnError);
   }
 }
 
@@ -69,8 +72,11 @@ class _ModuleRegistration extends Object with Disposable {
   _ModuleRegistration(this.module);
 
   @override
-  void manageStreamSubscription(StreamSubscription subscription) {
-    super.manageStreamSubscription(subscription);
+  StreamSubscription<T> listenToStream<T>(
+      Stream<T> stream, void onData(T event),
+      {Function onError, void onDone(), bool cancelOnError}) {
+    return super.listenToStream(stream, onData,
+        onError: onError, onDone: onDone, cancelOnError: cancelOnError);
   }
 }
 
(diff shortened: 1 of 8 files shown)


srinivasdhanwada-wf:
This on hold until new smithy runner image is available on drydock-prod

srinivasdhanwada-wf:
@Workiva/web-platform-pp Runner image has been updated, this is ready for review

lib/src/serializable.dart
```diff
@@ -138,6 +143,7 @@ class SerializableBus {
   void _deserializeAndCall(
       SerializableModule module, String method, List data) {
     InstanceMirror apiMirror = reflect(module.api);
+    List reflectedData = new List(data.length);
```
trentgrover-wf: why is this new list needed?

srinivasdhanwada-wf: When compiling with DDC, the original list `data` gets a runtime type of `List<Map<String, String>>`, which throws a runtime error when trying to assign a `JsonSerializable` to `data[i]`. I tried changing the type of `List<dynamic>` but that didn't work.

Adding the new list removes that type check and allows the list to be a `List<JsonSerializable>`

lib/src/serializable.dart
```diff
@@ -69,8 +72,11 @@ class _ModuleRegistration extends Object with Disposable {
   _ModuleRegistration(this.module);
 
   @override
-  void manageStreamSubscription(StreamSubscription subscription) {
-    super.manageStreamSubscription(subscription);
+  StreamSubscription<T> listenToStream<T>(
+      Stream<T> stream, void onData(T event),
+      {Function onError, void onDone(), bool cancelOnError}) {
+    return super.listenToStream(stream, onData,
+        onError: onError, onDone: onDone, cancelOnError: cancelOnError);
```
evanweible-wf: this seems to just be proxying directly to the super class with no changes – is an override actually needed?

lib/src/serializable.dart
evanweible-wf: Could we include the generic then?
```dart
final reflectedData = new List<JsonSerializable>(data.length);
```

lib/src/serializable.dart
srinivasdhanwada-wf: I don't think it's needed and will remove it.

lib/src/serializable.dart
srinivasdhanwada-wf: I'm not sure this would cover all cases. I think the reason my error was specific to `JsonSerializable` was because that was being used in the tests. The only requirement is that the class has a `.fromJson()` constructor. There could be a case where the class implements this constructor without extending/mixing in `JsonSerializable`.

We could add this generic and change the warning message to: "<className> must extend JsonSerializable"

maxwellpeterson-wf:
@jayudey-wf ready for you

smithy.yaml
```diff
@@ -1,11 +1,12 @@
...
 
-# dart 1.19.1
-runner_image: drydock-prod.workiva.org/workiva/smithy-runner-generator:92530
+# dart 1.24.2
+runner_image: drydock-prod.workiva.net/workiva/smithy-runner-generator:203768
 
 script:
-  - pub get
+  - pub get --packages-dir
+  - xvfb-run -s '-screen 0 1024x768x24' pub run dart_dev test --pub-serve --web-compiler=dartdevc -p chrome -p vm
```
jayudey-wf: does this need the vm flag here?

srinivasdhanwada-wf: It's not necessarily required, I added this during testing to point out errors that were specific to the DDC compilation (tests would pass in vm, but not in chrome). Since tests run fast in the VM, I don't think its a big deal to leave it in.

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - ran examples with ddc and verified that they were working as expected
  - verified that tests in smithy were being run with ddc as well
- [ ] All unit tests pass

Merging into master.

jayudey-wf:

rmconsole-wf:
+1 from RM

PR #100, opened 2017-08-22
WP-4919 unblock transition errors

toddbeckman-wf:
# Problem

If a module has an uncaught exception during an error, it is impossible to unload because it is in the middle of a transition.

# Solution

# Testing Suggestion

TBA


aviary2-wf:
## Raven
<details>

</details>

toddbeckman-wf:
Closing in favor of #103

PR #101, opened 2017-08-31
WP-4899 Make LifecycleModule a proper Disposable:

evanweible-wf:
## Changes
- `LifecycleModule` now extends `Disposable` instead of implementing the `DisposableManager` interface
- When unload() succeeds, the instance is also disposed
- When dispose() is called, it attempts to unload the module first, and then forces disposal regardless of the outcome of the unload step
- Prevents calls to all API methods after disposal has started

## Testing
- [ ] CI passes (tests updated & added)

## Code Review
@Workiva/web-platform-pp @Workiva/rich-app-platform-pp

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -17,7 +17,7 @@ library w_module.src.lifecycle_module;
 import 'dart:async';
 
 import 'package:logging/logging.dart';
-import 'package:meta/meta.dart' show protected, required;
+import 'package:meta/meta.dart' show mustCallSuper, protected, required;
 import 'package:w_common/disposable.dart';
 
 import 'package:w_module/src/simple_module.dart';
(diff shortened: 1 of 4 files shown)


aviary2-wf:
## Raven
<details>

</details>

lib/src/lifecycle_module.dart
```diff
@@ -600,8 +608,13 @@ abstract class LifecycleModule extends SimpleModule
     _state = LifecycleState.unloading;
     _transition = new Completer<Null>();
 
+    // Start the unload process.
     _unload(pendingTransition)
+        // If it succeeds, start the disposal.
+        .then((_) => dispose())
```
bencampbell-wf: 👍

georgelesica-wf:
@evanweible-wf you still working on this PR or is it superfluous?

evanweible-wf:
@georgelesica-wf it's just waiting for review

test/lifecycle_module_test.dart
```diff
@@ -948,20 +909,188 @@ void main() {
...
+              expectInLifecycleState(module, state);
+              await module.dispose();
+              expect(module.isDisposed, isTrue);
+              expect(
+                  module.eventList,
+                  equals([]
+                    ..addAll(expectedPreUnloadStates)
+                    ..addAll(['onShouldUnload', 'willUnload'])
+                    ..addAll(expectedDisposalStates)));
+            });
```
georgelesica-wf: Do we have a test somewhere that verifies that this works properly for child modules as well? I think we manage the child modules, so it is probably fine.

georgelesica-wf:
+1 with a question

evanweible-wf:
@Workiva/web-platform-pp @Workiva/rich-app-platform-pp ready for review.

George requested tests that exercise disposal when child modules are involved and that resulted in me catching a scenario where child modules that fail to unload were not then properly disposed when the parent module is disposed, which has now been fixed.

georgelesica-wf:
+1

test/lifecycle_module_test.dart
```diff
@@ -948,20 +909,263 @@ void main() {
...
-    test('getManagedPeriodicTimer should return a timer', () {
-      var callCount = 0;
-      module.getManagedPeriodicTimer(
-          new Duration(milliseconds: 10),
-          expectAsync1((Timer timer) {
-            if (callCount++ >= 1) {
-              timer.cancel();
-            }
-          }, count: 2));
+      void testDisposalFromLoadedState(LifecycleState state,
```
jayudey-wf: should this return a future that's `awaitable` since it's performing async operations in its body?

evanweible-wf: This method itself is not actually doing anything async (hence the `void` return type`) – it's simply defining a suite of tests via the `group()` and `test()` APIs based on the given parameters.

evanweible-wf:
@jayudey-wf ready for you

jayudey-wf:
noted issues in unit tests of consumers with this change

evanweible-wf:
The latest commit addresses a couple of race conditions that I discovered while working on datatable's tests.

The first issue is that a parent module listens to a child module's `didUnload` event and performs the following logic in response:

- calls the `onDidUnloadChildModule` handler (async)
- emits an event on the `didUnloadChildModule` stream
- cleans up state

However, the parent module was not previously waiting for this async handler to complete, which means that the parent module could be disposing or disposed by the time it completes. The result is a `StateError` when then trying to emit an event on the `didUnloadChildModule` stream because said stream has closed.

The solution is to use a `Completer` to monitor the status of each child module that is loaded. This completer is only completed when the child module unloads (either successfully or not), or when the module fails to load at all. We pass the `Completer.future` to `awaitBeforeDispose` in order to prevent the parent module from disposing until all of this child module logic has completed.

However, this change has a side effect: using `manageDisposable(childModule)` to force the disposal of child modules that fail to unload cleanly no longer works because that disposal logic will be blocked by the `awaitBeforeDispose()` tied to the child module (since those futures are awaited prior to executing any of the disposal logic).

The solution to this is to manually dispose of any child modules that are left after unloading the parent module. Note that to accomplish this, I changed the logic surrounding how the `_childModules` list is managed. Child modules are now only removed from the list when they actually unload instead of during their `willUnload` event. This required an additional boolean check around the `shouldUnload` logic in order to prevent checking a child module that is in the process of unloading.

You can manually confirm that this change was necessary by commenting out the following lines in `dispose()` and noting that several tests start failing due to a timeout exception:

```dart
    if (_childModules.isNotEmpty) {
      await Future.wait(_childModules.map((child) => child.dispose()));
    }
```

---

The last issue is that if the parent module is asked to unload while there are child modules in the process of loading, the parent module will not wait for these child modules to finish loading. This results in the parent module reaching a bad state due to it trying to unload/cleanup child modules while the relevant state (e.g. `_childModules`) is changing.

The solution is to keep track of child module loads via completers that are completed when the child module loads (either successfully or not). The futures from these completers are then awaited during the unload step of the parent module before actually proceeding with the unload logic. This ensures that all child modules are properly loaded such that they can then be properly unloaded.

Tests for this logic have been added.

lib/src/lifecycle_module.dart
```diff
@@ -264,18 +247,100 @@ abstract class LifecycleModule extends SimpleModule
...
-          {Function onError, void onDone(), bool cancelOnError}) =>
-      _disposableProxy.listenToStream(stream, onData,
-          onError: onError, onDone: onDone, cancelOnError: cancelOnError);
-
   //--------------------------------------------------------
   // Public methods that can be used directly to trigger
   // module lifecycle / check current lifecycle state
   //--------------------------------------------------------
 
+  /// Public method to dispose of this module.
```
bencampbell-wf: 🌵 This statement is more-or-less a rehash of the function signature. Maybe this would be more apt for a doc comment:

```diff
-  /// Public method to dispose of this module.
+  /// Disposes this module and all its disposable dependencies.
```

lib/src/lifecycle_module.dart
```diff
@@ -264,18 +247,100 @@ abstract class LifecycleModule extends SimpleModule
...
+    }
+
+    _customDisposal = new Completer<Null>();
+
+    if (isInstantiated) {
+      // If the module has not yet been loaded, then we can go straight to
+      // disposal.
+      _customDisposal.complete(super.dispose().then((_) => didDispose));
+      return _customDisposal.future;
+    }
```
bencampbell-wf: Testing if it is dispose or disposing and returning a single future for multiple calls to `dispose` is handled by the `super.dispose` [implementation](https://github.com/Workiva/w_common/blob/master/lib/src/common/disposable.dart#L323-L336). Why duplicate it here?

I may be dense and am just missing a critical detail.

```diff
   Future<Null> dispose() async {
-    if (isDisposedOrDisposing) {
-      return didDispose;
-    }
-
-    if (_customDisposal != null) {
-      return _customDisposal.future;
-    }
-
-    _customDisposal = new Completer<Null>();
-
     if (isInstantiated) {
-      // If the module has not yet been loaded, then we can go straight to
-      // disposal.
-      _customDisposal.complete(super.dispose().then((_) => didDispose));
-      return _customDisposal.future;
+      return super.dispose();
     }
```

lib/src/lifecycle_module.dart
```diff
@@ -264,18 +247,100 @@ abstract class LifecycleModule extends SimpleModule
...
+          stackTrace);
+    }
+
+    if (_childModules.isNotEmpty) {
+      await Future.wait(_childModules.map((child) => child.dispose()));
+    }
+
+    // Regardless of the unload outcome, dispose this instance as requested.
+    _customDisposal.complete(super.dispose().then((_) => didDispose));
+    return _customDisposal.future;
```
bencampbell-wf: This is similar to the comment above.

```diff
-    _customDisposal.complete(super.dispose().then((_) => didDispose));
-    return _customDisposal.future;
+    return super.dispose();
```

lib/src/lifecycle_module.dart
```diff
@@ -293,6 +358,10 @@ abstract class LifecycleModule extends SimpleModule
   /// Note that [LifecycleModule] only supports one load/unload cycle. If [load]
   /// is called after a module has been unloaded, a [StateError] is thrown.
   Future<Null> load() {
+    if (isDisposedOrDisposing || _customDisposal != null) {
```
bencampbell-wf: Is `_customeDisposal` redundant with respect to isDisposedOrDisposing?

lib/src/lifecycle_module.dart
evanweible-wf: I agree this would be much simpler, but unfortunately I have to handle some manual disposal logic before I can pass it off to `super.dispose()`. When `Module.dispose()` is called, we need to attempt to unload the module before moving on to disposing it. That unload process emits events which requires that `StreamController`s still be available. Due to the order in which disposal events happen in `Disposable.dispose()` (disposables are torn down and _then_ `onDispose()` is called), I have to override `dispose()` to do this instead of just overriding `onDispose()`. This also means that I need a separate piece to track this custom disposal phase since `isDisposedOrDisposing` won't be true until we actually call `super.dispose()` which doesn't happen until after attempting to unload the module.

I did quickly try moving the `onDispose()` step up in the `Disposable.dispose()` logic to see if I could then just override `onDispose()` and get rid of some of this custom logic, but I hit a bunch of stream-related errors so I would have to dig into that more to see if it's a workable solution.

bencampbell-wf: I see what you mean. I was hoping see duplicated Disposable logic removed 🤕

maxwellpeterson-wf:
For posterity:

> Evan Weible·17:38 The w_module PR can go in as is, I took another stab at the change Ben suggested yesterday but still didn’t have much luck so I’ll ping him and we can revisit it later

jayudey-wf:
any idea why the push build from travis CI is failing?

jayudey-wf:
installing this into a consuming library i am seeing test failures

maxwellpeterson-wf:
@jayudey-wf which library?

bencampbell-wf:
Travis appears to be trying to merge your PR branch into a target branch(b1d8df3_cr_no_smithy) that no longer exists instead of using master.

lib/src/lifecycle_module.dart
```diff
@@ -264,18 +245,47 @@ abstract class LifecycleModule extends SimpleModule
...
+  /// lifecycle handler like onUnload as a part of this disposal process, they
+  /// will still be available via their corresponding lifecycle event streams
+  /// (e.g. [didUnload]).
+  ///
+  /// The [Future] returned from this method will resolve when disposal has
+  /// completed and will only resolve with an error if one is thrown during
+  /// [onDispose].
+  @mustCallSuper
+  @override
+  Future<Null> dispose() => super.dispose();
```
maxwellpeterson-wf: Is this override still necessary? Is it just for mustCallSuper to work?

evanweible-wf: It's just so I can update the doc comment to explain how disposal works in tandem with unload.

maxwellpeterson-wf:
@jayudey-wf ready for you

test/lifecycle_module_test.dart
```diff
@@ -948,20 +908,265 @@ void main() {
...
+
+                test('should emit the unload failure from didUnload', () async {
+                  expect(module.didUnload.first, throwsA(same(testError)));
+                  if (withChild) {
+                    expect(
+                        childModule.didUnload.first, throwsA(same(testError)));
+                  }
+                  expectInLifecycleState(module, state);
+                  await module.dispose();
+                  print('DISPOSED');
```
jayudey-wf: #nit not needed anymore, could be cleaned up on next update. no need to re-review for it

jayudey-wf:
+1

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - verified in example that it was working as expected
  - installed into 4 consumers and verified that tests were still passing
- [ ] All unit tests pass

Merging into master.

jayudey-wf:

rmconsole-wf:
+1 from RM

PR #103, opened 2017-10-19
WP-4919 Unblock mid-transition states

toddbeckman-wf:
# Problem

If a module has an uncaught exception during an error, it is impossible to unload because it is in the middle of a transition.

# Solution

# Testing Suggestion

- [ ] CI passes

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -330,7 +330,6 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
 
     _load().then(transition.complete).catchError((error, trace) {
       transition.completeError(error, trace);
-      _state = LifecycleState.loaded;
     });
 
     return transition.future;
@@ -468,16 +467,16 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       pendingTransition = _transition.future;
     }
 
-    _transition = new Completer<Null>();
+    final transition = new Completer<Null>();
+    _transition = transition;
     _state = LifecycleState.suspending;
 
     _suspend(pendingTransition)
-        .then(_transition.complete)
+        .then(transition.complete)
         .catchError((error, trace) {
-      _transition.completeError(error, trace);
-      _state = LifecycleState.loaded;
+      transition.completeError(error, trace);
     });
-    return _transition.future;
+    return transition.future;
   }
 
   /// Public method to resume the module.
@@ -527,13 +526,13 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
     }
 
     _state = LifecycleState.resuming;
-    _transition = new Completer<Null>();
+    final transition = new Completer<Null>();
+    _transition = transition;
 
     _resume(pendingTransition)
-        .then(_transition.complete)
+        .then(transition.complete)
         .catchError((error, trace) {
-      _transition.completeError(error, trace);
-      _state = LifecycleState.suspended;
+      transition.completeError(error, trace);
     });
 
     return _transition.future;
@@ -626,11 +625,12 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
 
     _previousState = _state;
     _state = LifecycleState.unloading;
-    _transition = new Completer<Null>();
+    final transition = new Completer<Null>();
+    _transition = transition;
 
     var unloadAndDispose = new Completer<Null>();
-    unloadAndDispose.complete(_transition.future.then((_) => dispose()));
-    _transition.complete(_unload(pendingTransition));
+    unloadAndDispose.complete(transition.future.then((_) => dispose()));
+    transition.complete(_unload(pendingTransition));
     return unloadAndDispose.future;
   }
 
@@ -897,7 +897,6 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       // not explicitly cancel the unload), emit the unload failure event and
       // then rethrow the exception so that the caller (either unload() or
       // onWillDispose()) can handle it.
-      _state = LifecycleState.unloaded;
       _didUnloadController.addError(error, stackTrace);
       rethrow;
     }
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -897,6 +897,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       // not explicitly cancel the unload), emit the unload failure event and
       // then rethrow the exception so that the caller (either unload() or
       // onWillDispose()) can handle it.
+      _state = LifecycleState.unloaded;
       _didUnloadController.addError(error, stackTrace);
       rethrow;
     }
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -321,11 +321,19 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
     }
 
     _state = LifecycleState.loading;
-    _transition = new Completer<Null>();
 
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Raven
<details>

</details>

rmconsole-wf:
@toddbeckman-wf This pull request has merge conflicts, please resolve.

lib/src/lifecycle_module.dart
```diff
@@ -321,11 +321,19 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
     _state = LifecycleState.loading;
-    _transition = new Completer<Null>();
 
-    _load().then(_transition.complete).catchError(_transition.completeError);
+    // Keep track of this load's completer
+    final transition = new Completer<Null>();
 
-    return _transition.future;
+    // because this one can get overwritten
+    _transition = transition;
```
evanweible-wf: Do we need to do this for the other transitions as well? I'm not seeing anything that would make this only an issue for `load`

lib/src/lifecycle_module.dart
```diff
@@ -321,11 +321,19 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+
+    _load().then(transition.complete).catchError((error, trace) {
+      transition.completeError(error, trace);
+      _state = LifecycleState.loaded;
```
evanweible-wf: For suspend and resume, if the transition fails, the state is reverted. In other words..

- `loaded` to `isSuspending` back to `loaded`
- `suspended` to `isResuming` back to `suspended`

But for load and unload, we are forcing through to the expected state despite the failure..
- `initialized` to `isLoading` to `loaded`, despite `onLoad` failure
- `loaded` to `isUnloading` to `unloaded`, despite `onUnload` failure

It seems a bit odd that we would have differing (in fact, opposite) behaviors for these two sets

evanweible-wf:
I'd like to get some additional opinions from people who have worked on `Module` before. Are we okay with the general idea behind this change? Either reverting or forcing a transition in the event of a failure?

@maxwellpeterson-wf @todbachman-wf @georgelesica-wf @bencampbell-wf

todbachman-wf:
I'm skeptical. It doesn't seem clear to me that reverting to the previous state is always the right answer. Ideally, the module author would have retry logic in the methods that trigger these errors. In that case then an exception raised by one of these methods really should be treated exceptionally and we shouldn't try to recover.

toddbeckman-wf:
If a module fails an error transition, it is not safe to assume that forcing it into the next state, reverting to the previous state, or automatically unloading (and reloading?) the module are correct solutions. The only remaining option would be to do nothing about it, which would leave the module in a partial transition from which no further actions will function and the module remains broken. While I agree it is a rather silly decision to move it to the next state for two actions but to revert it in the other two, I don't see a clean solution.

evanweible-wf:
@toddbeckman-wf now that Modules are proper `Disposable`s, could we just make sure that a module can always be disposed, even if it's in a failed state like this?

todbachman-wf:
It seems to me that the right answer then is to leave it in a failed state and maybe surface that somehow?

lib/src/lifecycle_module.dart
toddbeckman-wf: Added.

lib/src/lifecycle_module.dart
```diff
@@ -321,11 +321,18 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
-    _load().then(_transition.complete).catchError(_transition.completeError);
+    // Keep track of this load's completer
+    final transition = new Completer<Null>();
 
-    return _transition.future;
+    // because this one can get overwritten
+    _transition = transition;
+
+    _load().then(transition.complete).catchError((error, trace) {
+      transition.completeError(error, trace);
```
evanweible-wf: Maybe there was something here before that necessitated the extra closure, I can't remember, but at this point we're just completing the transition with the result of `_load()`. I think you can go back to something like this:

```dart
_transition = new Completer<Null>();
_transition.complete(_load());
return _transition.future;
```

Since we're setting up the completion immediately, there's no need for the extra `transition` var – it doesn't matter if `_transition` gets overwritten later

lib/src/lifecycle_module.dart
```diff
@@ -454,14 +461,22 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
-    _suspend(previousTransition)
-        .then(_transition.complete)
-        .catchError(_transition.completeError);
-    return _transition.future;
+    _suspend(pendingTransition)
+        .then(transition.complete)
+        .catchError((error, trace) {
+      transition.completeError(error, trace);
+    });
+    return transition.future;
```
evanweible-wf: Similarly here, you can get around the need for the extra `transition` var if you setup the completion immediately:

```dart
var previousTransition = _transition?.future;
_transition = new Completer<Null>();
_state = LifecycleState.suspending;

_transition.complete(_suspend(previousTransition));
return _transition.future;
```

lib/src/lifecycle_module.dart
```diff
@@ -505,13 +520,20 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
-    _transition = new Completer<Null>();
+    final transition = new Completer<Null>();
+    _transition = transition;
 
     _resume(pendingTransition)
-        .then(_transition.complete)
-        .catchError(_transition.completeError);
+        .then(transition.complete)
+        .catchError((error, trace) {
+      transition.completeError(error, trace);
```
evanweible-wf: same thing here

lib/src/lifecycle_module.dart
```diff
@@ -596,14 +618,19 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
     _state = LifecycleState.unloading;
-    _transition = new Completer<Null>();
+    final transition = new Completer<Null>();
+    _transition = transition;
 
     var unloadAndDispose = new Completer<Null>();
-    unloadAndDispose.complete(_transition.future.then((_) => dispose()));
-    _transition.complete(_unload(pendingTransition));
+    unloadAndDispose.complete(transition.future.then((_) => dispose()));
+    transition.complete(_unload(pendingTransition));
```
evanweible-wf: This one was already setup in the way I suggested above, so again the extra `transition` var is not needed.

lib/src/lifecycle_module.dart
```diff
@@ -683,7 +710,10 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       if (isUnloading) {
         unloadingTransitionFuture = _transition.future;
       } else {
-        var pendingTransition = _transition?.future;
+        Future pendingTransition;
+        if (_transition != null && !_transition.isCompleted) {
+          pendingTransition = _transition?.future;
```
evanweible-wf: I'm wondering now if this is really necessary? If it's null, the handlers that these "pending transitions" get passed to already null-check, and if it's already completed, it's just effectively an `await new Future(() {});`

evanweible-wf: Ignore this, Todd pointed out that keeping a pending transition around indefinitely will just pass the error on whenever it is checked

greglittlefield-wf:
+1 to latest two commits

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - see above
- [ ] All unit tests pass
- [ ] Areas of regression noted (gathered from offline discussion)

Merging into master.

jayudey-wf:

PR #108, opened 2018-02-25
Replace dart:mirrors with package:reflectable

robbecker-wf:
# Problem
dart:mirrors is going away in Dart 2. It also results in large JS output, even with the @MirrorsUsed annotations.

# Solution
Replace it with the reflectable package

# Testing
 - [ ] Test with w_viewer_mobile and the mobile wdesk app ( cc @brianblanchard-wf )
 - [ ] Ensure no regressions in consumers (wdesk_sdk, product lines)

### DIFF
--- a/lib/src/serializable.dart
+++ b/lib/src/serializable.dart
@@ -26,10 +26,11 @@ import 'package:w_module/src/module.dart';
 
 // Any classes / methods that are going to be reflected must annotate with this
 class Reflectable extends reflect.Reflectable {
-  const Reflectable() :super(
-    reflect.invokingCapability, 
-    reflect.typingCapability,
-    );
+  const Reflectable()
+      : super(
+          reflect.invokingCapability,
+          reflect.typingCapability,
+        );
 }
 
 const _reflector = const Reflectable();
@@ -138,7 +139,6 @@ class SerializableBus {
 
   void _deserializeAndCall(
       SerializableModule module, String method, List data) {
-        
     reflect.InstanceMirror apiMirror = _reflector.reflect(module.api);
 
     if (apiMirror == null) {
@@ -167,12 +167,12 @@ class SerializableBus {
     for (var i = 0; i < apiMethodMirror.parameters.length; i++) {
       reflect.ParameterMirror param = apiMethodMirror.parameters[i];
       if (data[i] is Map && param.type.reflectedType != Map) {
-        reflect.ClassMirror paramClassMirror = _reflector.reflectType(param.type.reflectedType);
+        reflect.ClassMirror paramClassMirror =
+            _reflector.reflectType(param.type.reflectedType);
 
         // Paramter type must implement fromJson name constructor that takes a Map
         try {
-          var instance = paramClassMirror
-              .newInstance('fromJson', [data[i]]);
+          var instance = paramClassMirror.newInstance('fromJson', [data[i]]);
           reflectedData[i] = instance;
         } on NoSuchMethodError {
           _logger.warning(
--- a/lib/src/serializable.dart
+++ b/lib/src/serializable.dart
@@ -15,9 +15,8 @@
 library serializable_module.src.serializable;
 
 import 'dart:async';
-@MirrorsUsed(metaTargets: 'serializable_module.src.serializable.Reflectable')
-import 'dart:mirrors';
 
+import 'package:reflectable/reflectable.dart' as reflect;
 import 'package:logging/logging.dart';
 import 'package:w_common/disposable.dart' show Disposable;
 import 'package:w_common/json_serializable.dart' show JsonSerializable;
@@ -26,9 +25,14 @@ import 'package:w_module/src/event.dart';
 import 'package:w_module/src/module.dart';
 
 // Any classes / methods that are going to be reflected must annotate with this
-class Reflectable {
-  const Reflectable();
+class Reflectable extends reflect.Reflectable {
+  const Reflectable() :super(
+    reflect.invokingCapability, 
+    reflect.typingCapability,
+    );
 }
+
+const _reflector = const Reflectable();
 
 abstract class Bridge<T> extends Object with Disposable {
   final Stream<Map> apiCallReceived;
@@ -134,8 +138,8 @@ class SerializableBus {
 
   void _deserializeAndCall(
       SerializableModule module, String method, List data) {
-    InstanceMirror apiMirror = reflect(module.api);
-    List reflectedData = new List(data.length);
+        
+    reflect.InstanceMirror apiMirror = _reflector.reflect(module.api);
 
     if (apiMirror == null) {
       _logger.warning(
@@ -143,8 +147,8 @@ class SerializableBus {
       return;
     }
 
-    ClassMirror classMirror = apiMirror.type;
-    MethodMirror apiMethodMirror = classMirror.declarations[new Symbol(method)];
+    reflect.ClassMirror classMirror = apiMirror.type;
+    reflect.MethodMirror apiMethodMirror = classMirror.declarations[method];
 
     if (apiMethodMirror == null) {
       _logger.warning(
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Module Findings
No security relevant content detected. Please review for security relevance and request security review as needed.

test/serializable_test.dart
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-@TestOn('vm || browser')
```
robbecker-wf: when running these tests on the VM and the browser .. with the added reflectable transformer I get
```
 00:03 +199 -1: [VM] loading test/serializable_test.dart [E]
      Failed to load "test/serializable_test.dart": Unsupported operation: The transformed code is running with the untransformed reflectable package. Remember to set your package-root to 'build/.../packages'.
      http://localhost:62337/serializable_test.dart 28:5               _initializeReflectable
      http://localhost:62337/serializable_test.dart 18:3               main
      ===== asynchronous gap ===========================
      package:test                                                     internalBootstrapVmTest
      http://localhost:62337/serializable_test.dart.vm_test.dart 8:13  main
```
I couldn't figure out how to set the package-root, so just avoided the problem by not running the test on VM.

lib/src/serializable.dart
```diff
@@ -26,10 +25,16 @@ import 'package:w_module/src/event.dart';
 import 'package:w_module/src/module.dart';
 
 // Any classes / methods that are going to be reflected must annotate with this
-class Reflectable {
-  const Reflectable();
+class Reflectable extends reflect.Reflectable {
```
georgelesica-wf: What's the export for this file look like, this won't get exported, will it?

lib/src/serializable.dart
```diff
@@ -134,17 +139,16 @@ class SerializableBus {
 
   void _deserializeAndCall(
       SerializableModule module, String method, List data) {
-    InstanceMirror apiMirror = reflect(module.api);
-    List reflectedData = new List(data.length);
+    reflect.InstanceMirror apiMirror = _reflector.reflect(module.api);
```
georgelesica-wf: This type can't be inferred?

lib/src/serializable.dart
```diff
@@ -134,17 +139,16 @@ class SerializableBus {
...
     if (apiMirror == null) {
       _logger.warning(
           'Unable to create mirror on api for ${module.serializableKey}');
       return;
     }
 
-    ClassMirror classMirror = apiMirror.type;
-    MethodMirror apiMethodMirror = classMirror.declarations[new Symbol(method)];
+    reflect.ClassMirror classMirror = apiMirror.type;
+    reflect.MethodMirror apiMethodMirror = classMirror.declarations[method];
```
georgelesica-wf: What about these?

lib/src/serializable.dart
```diff
@@ -159,16 +163,16 @@ class SerializableBus {
       return;
     }
 
+    List reflectedData = new List(data.length);
```
georgelesica-wf: Can we do better on this generic type?

lib/src/serializable.dart
```diff
@@ -159,16 +163,16 @@ class SerializableBus {
...
     for (var i = 0; i < apiMethodMirror.parameters.length; i++) {
-      var param = apiMethodMirror.parameters[i];
-
+      reflect.ParameterMirror param = apiMethodMirror.parameters[i];
```
georgelesica-wf: Another place it seems like we should be able to simplify to `var`?

lib/src/serializable.dart
```diff
@@ -159,16 +163,16 @@ class SerializableBus {
...
       if (data[i] is Map && param.type.reflectedType != Map) {
-        ClassMirror paramClassMirror = reflectClass(param.type.reflectedType);
+        reflect.ClassMirror paramClassMirror =
```
georgelesica-wf: Another one...

lib/src/serializable.dart
robbecker-wf: List<dynamic> ?

georgelesica-wf:
Have you measured the new JS output size? Might be interesting to call that out on the PR?

lib/src/serializable.dart
jacehensley-wf: image

jacehensley-wf:
Will all consumers need to include the transformer?

```yaml
transformers:
- reflectable:
    entry_points:
      - web/main.dart # The path to your main file
```

lib/src/serializable.dart
georgelesica-wf: Hmm, I guess there aren't really types on the items that get put into the list, so that's fine.

lib/src/serializable.dart
robbecker-wf: It is exported intentionally. I made no changes to the public api here, just changed the `extends` here to use the reflectable package. 
FYI The `@Reflectable` annotation is used in w_viewer.

robbecker-wf:
> Will all consumers need to include the transformer?

@jacehensley-wf Normally, yes but the only consumer of serializable is w_viewer_mobile. So there's only 1 place to update.

robbecker-wf:
@georgelesica-wf I haven't been able to measure the difference yet, but since mirrors are being dropped for Dart 2 .. we'll need to get rid of them anyway. The size drop will just be a bonus.

lib/src/serializable.dart
georgelesica-wf: Oh OK, gotcha. Why the `_reflector` instance?

robbecker-wf: the _reflector instance is used further down to reflect on things. That's how you use it according to the docs https://github.com/dart-lang/reflectable/tree/master/reflectable#usage

georgelesica-wf: Is this still going to work then?

https://github.com/Workiva/w_viewer/blob/00e7ce49d522123eea0bc1b77bb6182b19f86441/lib/src/modules/content_renderer/content_renderer_api.dart#L22

robbecker-wf:
Well, this isn't great when trying to compile w_viewer
```
Error from Dart2JS]:
web/main.dart:
The compiler crashed when compiling this element.
The compiler is broken.

When compiling the above element, the compiler crashed. It is not
possible to tell if this is caused by a problem in your program or
not. Regardless, the compiler should not crash.

The Dart team would greatly appreciate if you would take a moment to
report this problem at http://dartbug.com/new.

Please include the following information:

* the name and version of your operating system,

* the Dart SDK build number (build number could not be determined), and

* the entire message you see here (including the full stack trace
  below as well as the source location above).

Build error:
Transform Dart2JS on w_viewer_app|web/main.dart threw error: The getter 'uses' was called on null.
Receiver: null
Tried calling: uses
dart:core                                               Object.noSuchMethod
package:compiler_unsupported/src/dump_info.dart 580     DumpInfoTask.dumpInfoJson
package:compiler_unsupported/src/dump_info.dart 540     DumpInfoTask.dumpInfo.<fn>
package:compiler_unsupported/src/common/tasks.dart 63   CompilerTask.measure
package:compiler_unsupported/src/dump_info.dart 537     DumpInfoTask.dumpInfo
package:compiler_unsupported/src/compiler.dart 629      Compiler.compileLoadedLibraries.<fn>
package:compiler_unsupported/src/common/tasks.dart 176  CompilerTask.measureSubtask
package:compiler_unsupported/src/compiler.dart 518      Compiler.compileLoadedLibraries
package:compiler_unsupported/src/compiler.dart 473      Compiler.runInternal
===== asynchronous gap ===========================
package:compiler_unsupported/src/compiler.dart 308      Compiler.run.<fn>.<fn>
dart:async                                              new Future.sync
package:compiler_unsupported/src/compiler.dart 308      Compiler.run.<fn>
package:compiler_unsupported/src/common/tasks.dart 176  CompilerTask.measureSubtask
package:compiler_unsupported/src/compiler.dart 305      Compiler.run
package:compiler_unsupported/src/apiimpl.dart 263       CompilerImpl.run.<fn>.<fn>

Build failed.
```

robbecker-wf:
I have to park this for a while unless someone else wants to get it over the finish line.

robbecker-wf:
Closing: I took a different approach of just getting rid of all reflection.

PR #110, opened 2018-05-02
Hide lifecycle logs by default 🤫

tomconnell-wf:
# Problem 
Module lifecycle logs are pretty noisy.  They are hard to filter, also.  If you filter out console logs that have either w_module or LifecycleModule in them, you wipe out many exceptions, because most of our ecosystem extends from LifeCycleModules.

# Solution
Keep error logging on, but hide normal lifecycle logs behind a url parameter, `w_module.verbose=true`

# Example
The amount of lifecycle logging before:
screen shot 2018-05-02 at 8 54 34 am

After (notice the length of the scrollbar pill):
image

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -18,7 +18,7 @@ import 'dart:async';
 
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart'
-    show mustCallSuper, protected, required, visibleForTesting;
+    show mustCallSuper, protected, required;
 import 'package:w_common/disposable.dart';
 
 import 'package:w_module/src/simple_module.dart';
@@ -53,7 +53,9 @@ enum LifecycleState {
 /// Intended to be extended by most base module classes in order to provide a
 /// unified lifecycle API.
 abstract class LifecycleModule extends SimpleModule with Disposable {
-  @visibleForTesting
+  /// Enables verbose logging for this module.
+  /// 
+  /// Globally overridden by the 'w_module.verbose' url parameter.
   bool verboseLogging = false;
   List<LifecycleModule> _childModules = [];
   Logger _logger;
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -17,7 +17,7 @@ library w_module.src.lifecycle_module;
 import 'dart:async';
 
 import 'package:logging/logging.dart';
-import 'package:meta/meta.dart' show mustCallSuper, protected, required;
+import 'package:meta/meta.dart' show mustCallSuper, protected, required, visibleForTesting;
 import 'package:w_common/disposable.dart';
 
 import 'package:w_module/src/simple_module.dart';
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/lifecycle_module.dart
```diff
@@ -791,7 +791,8 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
   /// A utility to logging LifecycleModule lifecycle events
   void _logLifecycleEvents(
       String logLabel, Stream<dynamic> lifecycleEventStream) {
-    listenToStream(lifecycleEventStream, (_) => _logger.fine(logLabel),
+
+    listenToStream(lifecycleEventStream, (_) { if (Uri.base.queryParameters['w_module.verbose'] == 'true') {_logger.fine(logLabel);}},
```
georgelesica-wf: Could we maybe do this as a global variable, sort of like what WSD does with its test mode, and then let applications determine how / when to set it?

patkujawa-wf: I like that idea too. Opt-in, right? e.g. `enableLifecycleLogging()`. What object would that live on?

lib/src/lifecycle_module.dart
```diff
@@ -52,6 +53,8 @@ enum LifecycleState {
 /// Intended to be extended by most base module classes in order to provide a
 /// unified lifecycle API.
 abstract class LifecycleModule extends SimpleModule with Disposable {
+  @visibleForTesting
```
georgelesica-wf: We probably don't even need this, someone might want to enable it this way in an app. Maybe make it private and module-level and then add a static or module-level function to flip it to true.

So, at the module level (or statically on the `LifecycleModule` class):

```dart
bool _verboseLogging = false;

void enableVerboseLogging() => _verboseLogging = true;
```

tomconnell-wf: Maybe both!

tomconnell-wf: I'll leave it per-object for now.

tomconnell-wf: I'm probably missing something subtle, but I'm not sure why the property would need to be private with a function to set it.

tomconnell-wf:
I considered that originally, but figured it would be better to not have everyone do it a different way.  Okay.

dustinlessard-wf:
@Workiva/app-frameworks , let's find a way to resolve this problem.  This is a very important concern that we can help alleviate.

Thanks to @tomconnell-wf for bringing this up.  We're going to prioritize this issue.

georgelesica-wf:
What about not using a URL parameter and just having a method to enable additional logging? Or disable it, whichever way people prefer. Then that's a lever that can be pulled easily by any app.

tomconnell-wf:
I'm open to any central method, and I get there could be concern that not having the lifecycle messages might make debugging harder in some cases.  

Maybe the bigger story (maybe an RFD) is how we handle overwhelming logs.  In any of the top-level app repos, there's a lot of logging built up over w_module, w_session, w_oauth2, etc, etc.  When I'm debugging the last mile of a bug, those logs are almost always just noise.  Filtering by log level isn't enough; I want to see some debug logs, but not the whole ecosystem's debug logs.  How do we draw the boundaries?  

Maybe I would not be as concerned with DDC when I can use a debugger again.

dustinlessard-wf:
We have an RFD in progress to capture issues with our logging facilities : https://docs.google.com/document/d/1_v2OpVk-7wRxlZThwi2bT1ZFBYnLdCjHuaFvGkaJIAo/edit

PR #114, opened 2018-07-12
AF-1656 - Tracing for lifecycle transitions

toddbeckman-wf:
# Description

Pretty much every module could benefit from added tracing, especially if it comes free to consumers.

# Changes

- If a `globalTracer()` is configured for the application, create a span for each of the lifecycle transitions and expose it to consumers who may wish to apply additional tags or use it for `childOf` and `followsFrom` references to other tags.
- Pass on the `SpanContext` from the transition span to any child modules so their transitions have a `childOf` link back to this one.

# Testing

Play with the example and note the console logging.

### DIFF
--- a/example/panel/modules/sample_tracer.dart
+++ b/example/panel/modules/sample_tracer.dart
@@ -1,0 +1,162 @@
+import 'dart:async';
+import 'package:opentracing/opentracing.dart';
+
+class SampleSpan implements Span {
+  static int _nextId = 0;
+  final int _id = _nextId++;
+
+  @override
+  final List<Reference> references;
+
+  @override
+  final Map<String, dynamic> tags;
+
+  @override
+  final List<LogData> logData = [];
+
+  @override
+  final String operationName;
+
+  @override
+  SpanContext context;
+
+  @override
+  final DateTime startTime;
+  DateTime _endTime;
+
+  Completer<Span> _whenFinished = new Completer<Span>();
+
+  SampleSpan(
+    this.operationName, {
+    SpanContext childOf,
+    this.references,
+    DateTime startTime,
+    Map<String, dynamic> tags,
+  })
+      : this.startTime = startTime ?? new DateTime.now(),
+        this.tags = tags ?? {} {
+    if (childOf != null) {
+      references.add(new Reference.childOf(childOf));
+    }
+    setTag('span.kind', 'client');
+
+    final parent = parentContext;
+    if (parent != null) {
+      this.context = new SpanContext(spanId: _id, traceId: parent.traceId);
+      this.context.baggage.addAll(parent.baggage);
+    } else {
+      this.context = new SpanContext(spanId: _id, traceId: _id);
+    }
+  }
+
+  @override
+  void addTags(Map<String, dynamic> newTags) => tags.addAll(newTags);
+
+  @override
+  Duration get duration => _endTime?.difference(startTime);
+
+  @override
+  DateTime get endTime => _endTime;
+
+  @override
+  void finish({DateTime finishTime}) {
+    if (_whenFinished == null) {
+      return;
+    }
+
+    _endTime = finishTime ?? new DateTime.now();
+    _whenFinished.complete(this);
+    _whenFinished = null;
+  }
+
+  @override
+  void log(String event, {dynamic payload, DateTime timestamp}) =>
+      logData.add(new LogData(timestamp ?? new DateTime.now(), event, payload));
+
+  @override
+  SpanContext get parentContext =>
+      references.isEmpty ? null : references.first.referencedContext;
+
+  @override
+  void setTag(String tagName, dynamic value) => tags[tagName] = value;
+
+  @override
+  Future<Span> get whenFinished => _whenFinished.future;
+
+  @override
+  String toString() {
+    final sb = new StringBuffer('SampleSpan(');
+    sb
+      ..writeln('traceId: ${context.traceId}')
+      ..writeln('spanId: ${context.spanId}')
+      ..writeln('operationName: $operationName')
+      ..writeln('tags: ${tags.toString()}')
+      ..writeln('startTime: ${startTime.toString()}');
+
+    if (_endTime != null) {
+      sb
+        ..writeln('endTime: ${endTime.toString()}')
+        ..writeln('duration: ${duration.toString()}');
+    }
+
+    if (logData.isNotEmpty) {
+      sb.writeln('logData: ${logData.toString()}');
+    }
+
+    if (references.isNotEmpty) {
+      final reference = references.first;
+      sb.writeln(
+          'reference: ${reference.referenceType} ${reference.referencedContext.spanId}');
+    }
+
+    sb.writeln(')');
+
+    return sb.toString();
+  }
+}
+
+class SampleTracer implements AbstractTracer {
+  @override
+  SampleSpan startSpan(
+    String operationName, {
+    SpanContext childOf,
+    List<Reference> references,
+    DateTime startTime,
+    Map<String, dynamic> tags,
+  }) {
+    return new SampleSpan(
+      operationName,
+      childOf: childOf,
+      references: references,
+      startTime: startTime,
+      tags: tags,
+    )
+      ..whenFinished.then((span) {
+        print(span.toString());
+      });
+  }
+
+  @override
+  Reference childOf(SpanContext context) => new Reference.childOf(context);
+
+  @override
+  Reference followsFrom(SpanContext context) =>
+      new Reference.followsFrom(context);
+
+  @override
+  SpanContext extract(String format, dynamic carrier) {
+    throw new UnimplementedError(
+        'Sample tracer for example purposes does not support advanced tracing behavior.');
+  }
+
+  @override
+  void inject(SpanContext spanContext, String format, dynamic carrier) {
+    throw new UnimplementedError(
+        'Sample tracer for example purposes does not support advanced tracing behavior.');
+  }
+
+  @override
+  Future<dynamic> flush() {
+    return null;
+  }
+}
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -53,6 +53,10 @@ enum LifecycleState {
 /// Intended to be extended by most base module classes in order to provide a
 /// unified lifecycle API.
 abstract class LifecycleModule extends SimpleModule with Disposable {
+  static int _nextId = 0;
+
+  int _instanceId = _nextId++;
+
   List<LifecycleModule> _childModules = [];
   Logger _logger;
   String _name;
@@ -151,7 +155,11 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       references.add(new Reference.childOf(_childOfContext));
     }
 
-    return tracer.startSpan(operationName, references: references);
+    return tracer.startSpan(operationName, references: references)
+      ..addTags({
+        'module.name': name,
+        'module.instanceId': _instanceId,
+      });
   }
 
   /// Name of the module for identification in exceptions and debug messages.
@@ -347,7 +355,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
           reason: 'A module can only be loaded once.');
     }
 
-    _span = _startTransitionSpan('load($name)');
+    _span = _startTransitionSpan('load_module');
 
     _state = LifecycleState.loading;
 
@@ -499,7 +507,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
           ]);
     }
 
-    _span = _startTransitionSpan('suspend($name)');
+    _span = _startTransitionSpan('suspend_module');
 
     Future pendingTransition;
     if (_transition != null && !_transition.isCompleted) {
@@ -569,7 +577,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       pendingTransition = _transition.future;
     }
 
-    _span = _startTransitionSpan('resume($name)');
+    _span = _startTransitionSpan('resume_module');
 
     _state = LifecycleState.resuming;
     final transition = new Completer<Null>();
@@ -668,7 +676,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
           ]);
     }
 
-    _span = _startTransitionSpan('unload($name)');
+    _span = _startTransitionSpan('unload_module');
 
     Future pendingTransition;
     if (_transition != null && !_transition.isCompleted) {
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -456,7 +456,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
 
       try {
         _childModules.add(childModule);
-        childModule._parentContext = _activeSpan?.context;
+        childModule._parentContext = _loadSpanContext;
 
         await childModule.load();
         await onDidLoadChildModule(childModule);
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -158,8 +158,9 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
   /// Builds a span that conditionally applies a followsFrom reference if this module
   /// was loaded by a parent module.
   ///
-  /// Returns `null` if this module does not override the [name] getter or if
-  /// if no globalTracer is configured.
+  /// Returns `null` if no globalTracer is configured, or if this module does
+  /// not override the [name] getter (as the default name becomes nonsensical
+  /// when compiled to js).
   Span _startTransitionSpan(String operationName) {
     if (name == _name) {
       return null;
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -59,7 +59,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
 
   List<LifecycleModule> _childModules = [];
   Logger _logger;
-  String _name;
+  String _defaultName;
   LifecycleState _previousState;
   LifecycleState _state = LifecycleState.instantiated;
   Completer<Null> _transition;
@@ -133,7 +133,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       'didUnload': didUnload,
     }.forEach(_logLifecycleEvents);
 
-    _name = 'LifecycleModule($runtimeType)';
+    _defaultName = 'LifecycleModule($runtimeType)';
   }
 
   /// If this module is in a transition state, this is the Span capturing the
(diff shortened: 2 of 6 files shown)


aviary-wf:
## Security Insights

## Action Items
---

georgelesica-wf:
lgtm!

lib/src/lifecycle_module.dart
```diff
@@ -320,6 +347,9 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
           reason: 'A module can only be loaded once.');
     }
 
+    // TODO childOf or followsFrom
```
toddbeckman-wf: This TODO is probably done.

evanweible-wf:
Code changes look good, just had a couple small nits.

lib/src/lifecycle_module.dart
```diff
@@ -127,6 +134,35 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
     }.forEach(_logLifecycleEvents);
   }
 
+  /// If this module is in a transition state, this is the Span capturing the
+  /// transition state.
```
greglittlefield-wf: This doc comment could use some example usage.

greglittlefield-wf:
Oh, and could you elaborate on the unload part?
> This does not add tracing for unload because those futures are way too sensitive to change.

test/test_tracer.dart
```diff
@@ -0,0 +1,166 @@
+import 'dart:async';
+import 'package:opentracing/opentracing.dart';
+
+class TestSpan implements Span {
```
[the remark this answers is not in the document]
toddbeckman-wf: BasicTracer is not OSS

lib/src/lifecycle_module.dart
```diff
@@ -464,7 +511,15 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
 
     Future pendingTransition;
     if (_transition != null && !_transition.isCompleted) {
-      pendingTransition = _transition.future;
+      pendingTransition = _transition.future
+          // Need to `catchError` before `whenComplete` to prevent errors from
+          // getting thrown here when they should be handled elsewhere
+          .catchError((_) {})
```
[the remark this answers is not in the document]
toddbeckman-wf: When `_transition` is first created, it has a `.catchError()` attached to handle it.

However, I want to add an additional `.whenComplete()` to this same future. If I didn't have the `.catchError`, it would simultaneously be caught by the other `.catchError()` and also throw in this code. I don't care about the error in `.catchError()` handler here because it's already handled. I just have it so I can call `.whenComplete()` without this thread throwing an uncaught exception.

lib/src/lifecycle_module.dart
```diff
@@ -849,9 +926,15 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
-        childSuspendFutures.add(child.suspend());
+        childSuspendFutures.add(new Future(() async {
+          child._childOfContext = _activeSpan?.context;
+          return child.suspend().whenComplete(() {
+            child._childOfContext = null;
+          });
+        }));
       }
       await Future.wait(childSuspendFutures);
+      assert(activeSpan != null);
```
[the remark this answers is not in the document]
toddbeckman-wf: Leftover from testing

lib/src/lifecycle_module.dart
```diff
@@ -127,6 +134,35 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+  SpanContext _childOfContext;
+
+  /// Builds a span that conditionally applies a followsFrom reference if this module
+  /// was loaded by a parent module.
+  ///
+  /// Returns `null` if no globalTracer is configured.
+  Span _startTransitionSpan(String operationName) {
+    final tracer = globalTracer();
+    if (tracer == null) {
+      return null;
```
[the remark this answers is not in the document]
toddbeckman-wf: This is correct. I do want to exist out of this code if there is no global tracer.

toddbeckman-wf:
I spent a whole day focused on just the unload logic, trying to figure out some combination of manipulating the many `Future`s that would give me a good start and end to the spans, but throws during unload/dispose would always break out onto threads they weren't supposed to, or the memory of the modules would be leaked, or the futures would not resolve in the order they should.

example/panel/modules/sample_tracer.dart
```diff
@@ -0,0 +1,162 @@
+import 'dart:async';
+import 'package:opentracing/opentracing.dart';
+
+class SampleSpan implements Span {
```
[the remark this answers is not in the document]
toddbeckman-wf: BasicTracer is not OSS

lib/src/lifecycle_module.dart
toddbeckman-wf: I added an example but I'm not sure if it's what you meant.

lib/src/lifecycle_module.dart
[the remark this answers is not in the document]
greglittlefield-wf: I'm still a little confused... Where is the other `catchError`?

Also, if you're using `whenComplete`, you shouldn't have to add a `catchError`, since it'll get called even if future completes with an error.

I'm just not sure why `catchError` is needed here when it wasn't before these changes.

lib/src/lifecycle_module.dart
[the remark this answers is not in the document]
greglittlefield-wf: I don't think my [previous comment](https://github.com/Workiva/w_module/pull/114#discussion_r208409950) was fully addressed:
> Shouldn't this be null in cases where there's no globaTracer defined?
> ...
> Or, are we to assume globalTracer() is always non-null, and there's at least a NoopTracer available?
>
> So either:
>
> If the spans will sometimes be null, there should be test coverage for those cases (e.g., running tests with and without a global tracer set up)
> If the spans will never be null, we can remove all of the asserts or null-awares.

lib/src/lifecycle_module.dart
```diff
@@ -849,7 +939,12 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       _willSuspendController.add(this);
       List<Future<Null>> childSuspendFutures = <Future<Null>>[];
       for (var child in _childModules.toList()) {
-        childSuspendFutures.add(child.suspend());
+        childSuspendFutures.add(new Future(() async {
```
greglittlefield-wf: Sorry, forgot to comment on this one as well:

>#nit to help avoid race conditions in Dart 1, this could be new Future.sync

greglittlefield-wf:
Hm, did you try something similar to the following changes? If so, what issues did that approach have?
```diff
   Future<Null> _unload(Future<Null> pendingTransition) async {
     try {
       if (pendingTransition != null) {
         await pendingTransition;
       }
+      _activeSpan = _startTransitionSpan('unload');
 
       final shouldUnloadResult = shouldUnload();
       if (!shouldUnloadResult.shouldUnload) {
         _state = _previousState;
         _previousState = null;
         _transition = null;
         // reject with shouldUnload messages
         throw new ModuleUnloadCanceledException(
             shouldUnloadResult.messagesAsString());
       }
       _willUnloadController.add(this);
       await Future.wait(_childModules.toList().map((child) {
         child._parentContext = _activeSpan?.context;
         return child.unload().whenComplete(() {
           child._parentContext = null;
         });
       }));
       await onUnload();
       if (_state == LifecycleState.unloading) {
         _state = LifecycleState.unloaded;
         _previousState = null;
         _transition = null;
       }

       _didUnloadController.add(this);
     } on ModuleUnloadCanceledException catch (error, _) {
       // In the event of a cancellation, rethrow the exception and allow the
       // caller (either unload() or onWillDispose()) to handle it.
       rethrow;
     } catch (error, stackTrace) {
       // In the event of a failed unload (the module threw an exception but did
       // not explicitly cancel the unload), emit the unload failure event and
       // then rethrow the exception so that the caller (either unload() or
       // onWillDispose()) can handle it.
       _didUnloadController.addError(error, stackTrace);
+
+      _activeSpan?.setTag('error', true);

       rethrow;
+    } finally {
+      _activeSpan?.finish();
     }
```

greglittlefield-wf:
@toddbeckman-wf Might also be good to sync up once you get back on the stuff I've been working on around passing span references through Zone `zoneValues`, in case it comes in handy here.

lib/src/lifecycle_module.dart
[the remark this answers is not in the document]
toddbeckman-wf: The other catchError is in the previous transition's `await _transition`

Yes, `whenComplete` will get called, but the Future still throws synchronously on line 514.

lib/src/lifecycle_module.dart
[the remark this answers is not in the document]
toddbeckman-wf: This is the same testing issue as the other PR.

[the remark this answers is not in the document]
toddbeckman-wf: I added tests for this.

test/lifecycle_module_test.dart
```diff
@@ -277,19 +295,53 @@ Future<Null> executeStateTransition(
...
+        assert(globalTracer() == null);
+      });
+
+      runTests(false);
+    });
+  });
+}
+
+/// Returns the `globalTracer()` typecasted as a `TestTracer` or `null`.
+TestTracer globalTestTracer() {
```
toddbeckman-wf: Not my favorite name.

test/lifecycle_module_test.dart
```diff
@@ -1491,6 +1784,27 @@ void main() {
...
+            subs.add(globalTestTracer()
+                .onSpanFinish
+                .where((span) => span.tags['module.name'] == 'child')
+                .listen(expectAsync1(childSpanCompleter.complete)));
+
+            expect(parentModule.resume(),
+                throwsA(same(childModule.onResumeError)));
+
+            final span = await childSpanCompleter.future;
+            await new Future(() {}); // wait for parent to finish suspending
```
toddbeckman-wf: TODO comment should be resuming

lib/src/lifecycle_module.dart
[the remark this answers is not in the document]
dustinlessard-wf: > Or, are we to assume globalTracer() is always non-null. 

This is a safe assumption because by default, a NoopTracer is used.

lib/src/lifecycle_module.dart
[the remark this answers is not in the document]
greglittlefield-wf: Sorry, I'm a little confused... Could you point out exactly where that is?

If there is a catchError there, can we instead not return the result of `catchError` to ensure it's handled by the other `catchError` and not swallowed here? 

```dart
pendingTransition = _transition.future;
_transition.future.catchError((_) {}).whenComplete(() {
  _activeSpan = _startTransitionSpan('suspend_module');
});
```

Or, if something else is calling `catchError` on the returned future and not `_transition.future`, we could also use the result of `whenComplete` without a `catchError`.

```dart
pendingTransition = _transition.future.whenComplete(() {
  _activeSpan = _startTransitionSpan('suspend_module');
});
```

lib/src/lifecycle_module.dart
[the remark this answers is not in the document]
greglittlefield-wf: Should we remove null-awares and tests for this, then?

greglittlefield-wf:
One last comment around the `catchError`, otherwise LGTM

test/lifecycle_module_test.dart
```diff
@@ -2330,4 +2335,30 @@ void runTests(bool runSpanTests) {
...
+        final sub = getTestTracer().onSpanFinish.listen((_) {
+          fail('No span should be created');
+        });
+        addTearDown(sub.cancel);
+
+        await module.load();
+        await module.suspend();
+        await module.resume();
+        await module.unload();
+        await new Future.delayed(new Duration(milliseconds: 10));
```
greglittlefield-wf: What is this Future for? Could use a comment

lib/src/lifecycle_module.dart
```diff
@@ -125,11 +132,60 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+  Span get activeSpan => _activeSpan;
+
+  /// Set internally by the parent module if this module is called by [loadChildModule]
+  SpanContext _parentContext;
+
+  /// Builds a span that conditionally applies a followsFrom reference if this module
+  /// was loaded by a parent module.
+  ///
+  /// Returns `null` if this module does not override the [name] getter or if
+  /// if no globalTracer is configured.
```
greglittlefield-wf: Could we update this comment to say why? (tracing is disabled for unnamed modules since the default name gets mangled in dart2js)

test/lifecycle_module_test.dart
```diff
@@ -711,6 +793,66 @@ void main() {
         await module.suspend();
       });
 
+      if (runSpanTests) {
+        test('should record a span', () async {
+          await gotoState(module, LifecycleState.loaded);
+          subs.add(getTestTracer()
+              .onSpanFinish
+              .where((span) => span.operationName == 'LifecycleModule.suspend')
+              .listen((span) {
```
greglittlefield-wf: I think this is missing an `expectAsync1`

test/lifecycle_module_test.dart
```diff
@@ -711,6 +793,66 @@ void main() {
...
+          await module.suspend();
+        });
+
+        test(
+            'if a transition is in progress should wait to start this until that finishes',
+            () async {
+          await gotoState(module, LifecycleState.suspended);
+
+          Completer<Span> suspendCompleter = new Completer(),
+              resumeCompleter = new Completer();
```
greglittlefield-wf: #nit can we split this into two variables?

test/lifecycle_module_test.dart
```diff
@@ -1264,19 +1515,40 @@ void main() {
...
       parentModule = new TestLifecycleModule(name: 'parent');
       childModule = new TestLifecycleModule(name: 'child');
+
+      if (runSpanTests) {
+        subs.add(getTestTracer()
+            .onSpanFinish
+            .where((span) =>
+                span.operationName == 'LifecycleModule.load' &&
+                span.tags['module.name'] == 'parent')
+            .listen((span) {
```
greglittlefield-wf: Should this be wrapped in an `expectAsync1`?

test/lifecycle_module_test.dart
```diff
@@ -990,6 +1062,46 @@ void runTests(bool runSpanTests) {
...
+          await module.resume();
+
+          final suspendSpan = await suspendCompleter.future;
+          final resumeSpan = await resumeCompleter.future;
+
+          final suspendEnd = suspendSpan.startTime.add(suspendSpan.duration);
+
+          // checks that suspend ended before or at the same time the resume started
+          expect(
+              suspendEnd.compareTo(resumeSpan.startTime), lessThanOrEqualTo(0));
```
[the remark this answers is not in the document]
toddbeckman-wf:
```
 Expected: a value less than or equal to DateTime:<2018-09-05 08:53:15.174793>
        Actual: DateTime:<2018-09-05 08:53:15.174661>
         Which: is not a value less than or equal to DateTime:<2018-09-05 08:53:15.174793>
      NoSuchMethodError: Class 'DateTime' has no instance method '<'.
      Receiver: Instance of 'DateTime'
```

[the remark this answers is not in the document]
greglittlefield-wf: Ah, my bad!

greglittlefield-wf:
+10

lib/src/lifecycle_module.dart
```diff
@@ -52,12 +53,18 @@ enum LifecycleState {
 /// Intended to be extended by most base module classes in order to provide a
 /// unified lifecycle API.
 abstract class LifecycleModule extends SimpleModule with Disposable {
+  static int _nextId = 0;
+
+  int _instanceId = _nextId++;
+
   List<LifecycleModule> _childModules = [];
   Logger _logger;
   String _name;
```
evanweible-wf: I think I'd prefer renaming this to `_defaultName` so that the check you do to see if the module author has overridden the name or not makes more sense:

```dart
if (name == _defaultName) { ... }
```

lib/src/lifecycle_module.dart
```diff
@@ -475,7 +547,12 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
         .then(transition.complete)
         .catchError((error, trace) {
       transition.completeError(error, trace);
+      _activeSpan?.setTag('error', true);
```
evanweible-wf: Is there any value to attaching the stack trace to the span as well?

lib/src/lifecycle_module.dart
```diff
@@ -826,7 +911,12 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       _willResumeController.add(this);
       List<Future<Null>> childResumeFutures = <Future<Null>>[];
       for (var child in _childModules.toList()) {
-        childResumeFutures.add(child.resume());
+        childResumeFutures.add(new Future.sync(() {
+          child._parentContext = _activeSpan?.context;
+          return child.resume().whenComplete(() {
+            child._parentContext = null;
```
evanweible-wf: I'm a little confused by this. Wouldn't the parent-child relationship be valid regardless of what state the modules are in? Why do we need to set and then unset the `_parentContext` reference here?

toddbeckman-wf: The parent context refers to the current span. The parent's load context will be different from the parent's resume context, so if you just pass in one context and be done, you'll get every child's transition on the same parent's event.

lib/src/lifecycle_module.dart
toddbeckman-wf: There might be although I'd imagine the logging infrastructure is better for the job, and we're already logging these errors.

test/lifecycle_module_test.dart
```diff
@@ -1855,4 +2336,31 @@ void main() {
...
+        });
+        addTearDown(sub.cancel);
+
+        await module.load();
+        await module.suspend();
+        await module.resume();
+        await module.unload();
+        // Wait some time after transitions are over to be sure no spans are created
+        await new Future.delayed(new Duration(milliseconds: 10));
+      });
```
greglittlefield-wf: Sorry, saw one more thing while reviewing your dependent PR...

Can we also add a test for an unnamed module that has loaded a child module, to assert all the null-checks on the `_activeSpan` are working as intended?

toddbeckman-wf: Added

test/lifecycle_module_test.dart
```diff
@@ -2309,6 +2315,148 @@ void runTests(bool runSpanTests) {
...
+
+    tearDown(() async {
+      await parentModule.dispose();
+      await childModule.tearDown();
+      subs.forEach((sub) => sub.cancel());
+      subs.clear();
+    });
+
+    if (runSpanTests) {
+      test('should record a span for child with no parent', () async {
```
greglittlefield-wf: > for child with no parent

#nit "parent" is ambiguous here. What about something like:
> for child with unnamed parent, and thus no parentContext

toddbeckman-wf: I can fix this in 1540 which is also in CR right after this one

greglittlefield-wf:
+1

greglittlefield-wf:
QA +1

PR #115, opened 2018-07-31
AF-1540 AF-2674  consumer-specified timing for "first useful state"

toddbeckman-wf:
# Note

Merges on #114 . Clean diff: https://github.com/Workiva/w_module/compare/tracing_for_lifecycle_transitions...AF-1540_consumer-specified-timing

## Description

Allow consumers to specify when the module is actually in a useful state.

## Testing

See the `DataLoadAsync` example and verify that the `module_entered_first_useful_state` span is created after the data is loaded. It should have a `followsFrom` reference to the `load_module` span with the same `startTime` and `module.instanceId`.


aviary-wf:
## Security Insights

## Action Items
---

lib/src/timing_specifiers.dart
```diff
@@ -0,0 +1,16 @@
+class StartupTimingSpecifier {
```
toddbeckman-wf: Not my favorite name.

evanweible-wf:
I'm a little concerned with how this overlaps with the existing idleState event from Experience, but as long as we have a plan and clear documentation on what should be used (and maybe deprecate idle state?), then I'm okay with this

example/panel/modules/data_load_async_module.dart
```diff
@@ -42,8 +44,10 @@ class DataLoadAsyncModule extends Module {
   @protected
   Future<Null> onLoad() {
     // trigger non-blocking async load of data
+    _events.didLoadData.first
+        .then((_) => specifyStartupTiming(StartupTimingSpecifier.firstUseful));
     _actions.loadData();
-    return new Future.value();
+    return null;
```
greglittlefield-wf: Why is this returning null instead of a future?

lib/src/lifecycle_module.dart
```diff
@@ -52,12 +54,21 @@ enum LifecycleState {
 /// Intended to be extended by most base module classes in order to provide a
 /// unified lifecycle API.
 abstract class LifecycleModule extends SimpleModule with Disposable {
+  static int _nextId = 0;
+  // Used by tracing to tell apart multiple instances of the same module
```
greglittlefield-wf: Looks like this and other comments need to make their way into the lifecycle tracing PR

lib/src/lifecycle_module.dart
```diff
@@ -127,10 +138,84 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+    final tracer = globalTracer();
+    if (tracer == null) {
+      return null;
+    }
+
+    tracer
+        .startSpan(
+          specifier.name,
+          references: [tracer.followsFrom(_loadContext)],
+          startTime: _startLoadTime,
```
greglittlefield-wf: Would be nice if we started this span at the same time as load as opposed to passing in the start time so that we could tie spans into browser marks/measures

lib/src/timing_specifiers.dart
greglittlefield-wf: I was thinking "mark" or "indicate" for the verb for use in the method name, and then this could be "type" or un-suffixed or something.

```dart
markStartupTiming(StartupTiming.firstEditable);
indicateStartupTiming(StartupTimingType.firstEditable);
```

lib/src/timing_specifiers.dart
```diff
@@ -0,0 +1,16 @@
...
+  static const StartupTimingSpecifier firstComponentRender =
+      const StartupTimingSpecifier._('module_first_component_rendered');
+
+  static const StartupTimingSpecifier firstEditable =
+      const StartupTimingSpecifier._('module_entered_first_editable_state');
+
+  static const StartupTimingSpecifier firstReadable =
+      const StartupTimingSpecifier._('module_entered_first_readable_state');
+
+  static const StartupTimingSpecifier firstUseful =
```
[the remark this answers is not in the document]
greglittlefield-wf: @evanweible-wf Yeah, I think the intention is to unify or replace that event with this API.

+1 we should have pretty solid use-cases and guidelines for each type before committing to adding them. 

Thoughts on starting with just a constant for the equivalent to `onEnterIdleState` (i.e., AF-1540 and not AF-1541).

lib/src/lifecycle_module.dart
```diff
@@ -127,10 +138,84 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+    }
+
+    return tracer.startSpan(
+      operationName,
+      references: references,
+      tags: _defaultTags,
+    );
+  }
+
+  /// Creates a span from the start of [load] until now.
```
greglittlefield-wf: #nit for more context, it'd be nice to mention tracing here:

> Creates a span using [globalTracer]

or

> Creates a tracing span

lib/src/lifecycle_module.dart
```diff
@@ -127,10 +138,84 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+  /// to populate its store) will be finished by this point, whereas it might
+  /// only be started in [onLoad].
+  ///
+  /// Any [tags] specified will be added to the span which results from this call.
+  @protected
+  void specifyStartupTiming(
+    StartupTimingSpecifier specifier, {
+    Map<String, dynamic> tags: const {},
+  }) {
+    // Load didn't start
```
greglittlefield-wf: Should this throw a `StateError` instead of returning null?

lib/src/lifecycle_module.dart
```diff
@@ -127,10 +138,84 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+  /// This is different from [onLoad] as this indicates that whichever asynchronous
+  /// processes required to make this module interactible (such as any data needed
+  /// to populate its store) will be finished by this point, whereas it might
+  /// only be started in [onLoad].
+  ///
+  /// Any [tags] specified will be added to the span which results from this call.
+  @protected
+  void specifyStartupTiming(
+    StartupTimingSpecifier specifier, {
+    Map<String, dynamic> tags: const {},
```
greglittlefield-wf: Can we allow passing in other `references` so this reference can be associated with whatever triggered it, if that thing is traced?

lib/src/timing_specifiers.dart
greglittlefield-wf: This class and its constants need doc comments

test/lifecycle_module_test.dart
```diff
@@ -281,15 +299,22 @@ void main() {
   Logger.root.level = Level.ALL;
   final StateError testError = new StateError('You should have expected this');
 
+  final TestTracer tracer = new TestTracer();
+  initGlobalTracer(tracer);
+  assert(globalTracer() == tracer);
```
greglittlefield-wf: Same comments as in the last PR around testing and whether a non-null globalTracer is required

greglittlefield-wf: Same comments as in the last PR around testing and whether a non-null globalTracer is required

lib/src/timing_specifiers.dart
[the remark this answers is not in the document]
toddbeckman-wf: +1 to replacing that event.

The problem with just going with one option now and moving on is introducing a public API that we might change once we support other events. That was the main motivation to combining the PR's. I can undo it easily, but we should keep that in mind.

lib/src/lifecycle_module.dart
toddbeckman-wf: Then we would end up with unfinished spans for all modules that don't implement closing these.

lib/src/lifecycle_module.dart
toddbeckman-wf: There's already `followsFrom` reference to the load span.

lib/src/lifecycle_module.dart
toddbeckman-wf: Is it alright that this PR is already merging them?

greglittlefield-wf: It would be nice if they were in the other PR so that work could be reviewed atomically, but it's fine it it's too much effort to make that happen.

lib/src/lifecycle_module.dart
greglittlefield-wf: Good point, didn't think about that...

...Is it a problem to have unfinished spans?

toddbeckman-wf: Each unfinished span leaks a `Completer` in the span object and a `whenFinished` stream listener in the tracer (App Intelligence for us). A few might go unnoticed, but I'm not comfortable with the possibility that this will leak memory over time if we're automatically starting them for every module and only finishing them for the handful that actually implement stopping these.

lib/src/timing_specifiers.dart
```diff
@@ -0,0 +1,16 @@
+class StartupTimingType {
```
greglittlefield-wf: This comment got collapsed without being addressed:

> This class and its constants need doc comments.

lib/src/lifecycle_module.dart
```diff
@@ -127,10 +138,96 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+  ///
+  /// This is different from [onLoad] as this indicates that whichever asynchronous
+  /// processes required to make this module interactible (such as any data needed
+  /// to populate its store) will be finished by this point, whereas it might
+  /// only be started in [onLoad].
+  ///
+  /// Any [tags] specified will be added to the span which results from this call.
+  @protected
+  void specifyStartupTiming(
+    StartupTimingType specifier, {
```
greglittlefield-wf: Also got collapsed:

> ###### greglittlefield-wf 20 days ago  Member
> Can we allow passing in other `references` so this reference can be associated with whatever triggered it, if that thing is traced?
>
> #### toddbeckman-wf 19 days ago  Member
> There's already followsFrom reference to the load span.

Right, but there may be another relevant reference that could be specified. E.g., a document loads due to a traced network call returning, which triggers a call to `specifyStartupTiming`.

lib/src/lifecycle_module.dart
```diff
@@ -127,10 +138,96 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+    Map<String, dynamic> tags: const {},
+  }) {
+    // Load didn't start
+    if (_loadContext == null || _startLoadTime == null) {
+      return null;
```
greglittlefield-wf: This comment got collapsed without getting addressed:

>Should this throw a `StateError` instead of returning null?

lib/src/timing_specifiers.dart
```diff
@@ -0,0 +1,16 @@
...
+  final String name;
+  const StartupTimingType._(this.name);
+
+  static const StartupTimingType firstComponentRender =
```
greglittlefield-wf: Continuing this unresolved conversation: https://github.com/Workiva/w_module/pull/115#discussion_r208044080

> The problem with just going with one option now and moving on is introducing a public API that we might change once we support other events. That was the main motivation to combining the PR's. I can undo it easily, but we should keep that in mind.

I don't think that removing all but one constant vs leaving them as-is puts us at any more risk for changing the public API.

toddbeckman-wf: This comment was in regards to having it be a specific 'useful state' method call versus this enumeration approach.

toddbeckman-wf: It's pretty trivial, if we go this enum route, to have just one vs all of them. I don't see the advantage of splitting them up into multiple tickets.

lib/src/timing_specifiers.dart
toddbeckman-wf: Added

lib/src/lifecycle_module.dart
toddbeckman-wf: Added

toddbeckman-wf:
Continuing the conversation here so it doesn't get collapsed in the code changes:

I am okay with starting with just one. The "first useful" state was the one we originally discussed in Hipchat as the first meaningful one, with AF-1541 giving more optional granularity, so that's the one I started with for now. It's easy enough to refactor to whichever other one we ant if we change our minds.

lib/src/lifecycle_module.dart
```diff
@@ -176,18 +184,75 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+
+  /// Creates a span with `globalTracer` from the start of [load] until now.
+  ///
+  /// This span is intended to represent the time it takes for the module to
+  /// finish asynchronously loading any necessary data and entering a state which
+  /// is ready for user interaction.
+  ///
+  /// Any [tags] or [references] specified will be added to this span.
+  @protected
+  void specifyEnterFirstUsefulStateTiming({
```
toddbeckman-wf: Supercalifragilisticexpialidocious

greglittlefield-wf: Do we want to have shorthands for each one, or should we just have consumers use `specifyStartupTiming(StartupTimingType.firstUseful)`?

toddbeckman-wf:
Github doesn't let me respond to that comment so I'll just do it here

We can have shorthands if we want. I removed the others from this PR so there's nothing else to have shorthands for. However I'd be down for shortening the name by a few hundred characters. Kinda pointless shorthand if it's longer than without lol

test/lifecycle_module_test.dart
```diff
@@ -409,6 +436,60 @@ void runTests(bool runSpanTests) {
...
+                () async {
+              subs.add(getTestTracer()
+                  .onSpanFinish
+                  .where((span) =>
+                      span.operationName ==
+                      'TestLifecycleModule.${specifier.operationName}')
+                  .listen(expectAsync1((span) {
+                expect(span.startTime, startTime);
+              })));
+              module.specifyStartupTiming(specifier);
```
greglittlefield-wf: We should have test coverage for custom and default tags/references passed into this. (same for any shorthand methods we have).

greglittlefield-wf:
Sorry, in that conversation I was suggesting just removing the shorthand.

smithy.yaml
```diff
@@ -1,8 +1,7 @@
 project: dart
 language: dart
 
-# dart 1.24.2
-runner_image: drydock-prod.workiva.net/workiva/smithy-runner-generator:203768
+runner_image: google/dart:1.24.3
```
josephswanson-wf: may as well delete this file.. everything has been building on Workiva Build via the Dockerfile for quite some time
(I suspect Rosie is failing her dependency scan because of this not being an *allowed* Smithy runner image)

example/panel/modules/data_load_async_module.dart
```diff
@@ -42,6 +44,8 @@ class DataLoadAsyncModule extends Module {
   @protected
   Future<Null> onLoad() {
     // trigger non-blocking async load of data
+    listenToStream(_events.didLoadData.take(1),
+        (_) => specifyStartupTiming(StartupTimingType.firstUseful));
```
evanweible-wf: I don't think it really matters, but just for my own clarification, this could just be `specifyFirstUsefulState()` right?

toddbeckman-wf: Yep, either way works. We weren't decided on whether we wanted shorthand at the time the example was written.

documentation/tracing.md
```diff
@@ -0,0 +1,73 @@
...
+
+Note that `activeSpan` will be null at times when the module is not in the middle of a lifecycle transition.
+
+### Additional Load Time Granularity
+
+Sometimes, lifecycle methods such as `load` will complete before the module is semantically "loaded". For example, you may begin asynchronously fetching data for your module and then return from `onLoad` to keep from blocking the main thread.
+
+In cases like these, use `specifyStartupTiming`:
+
+```
```
greglittlefield-wf: #nit could be a Dart code block.

lib/src/lifecycle_module.dart
```diff
@@ -177,16 +185,74 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
+    _startLoadTime = null;
   }
 
   /// Name of the module for identification in exceptions and debug messages.
   // ignore: unnecessary_getters_setters
   String get name => _defaultName;
 
+  Map<String, dynamic> get _defaultTags => {
+        'span.kind': 'client',
+        'module.instanceId': _instanceId,
```
greglittlefield-wf: #not I know this wasn't changed in this PR, but should this be `.instance_id`, not `.instanceId`?

example/panel/modules/data_load_async_module.dart
```diff
@@ -62,15 +66,20 @@ class DataLoadAsyncActions {
...
+
+class DataLoadAsyncEvents {
+  final Event didLoadData = new Event(_dispatchKey);
+}
+
 class DataLoadAsyncStore extends Store {
   DataLoadAsyncActions _actions;
-  List<String> _data;
-  bool _isLoading;
+  DataLoadAsyncEvents _events;
```
greglittlefield-wf: #nit this is never disposed

toddbeckman-wf: Added

pubspec.yaml
```diff
@@ -13,16 +13,16 @@ homepage: https://github.com/Workiva/w_module
...
   w_common: ^1.9.0
 
 dev_dependencies:
   browser: ^0.10.0+2
   coverage: ^0.7.3
   dart_dev: ^1.8.0
-  dart_style: ^0.2.16
-  dartdoc: ^0.9.0
+  dart_style: ^1.0.7
+  dartdoc: ">=0.13.0 <0.16.0"
```
toddbeckman-wf: Copied from wdesk_sdk. Shouldn't affect OSS consumers since it's a dev dep

PR #120, opened 2018-08-16
AF-1777 Module name upgrader

toddbeckman-wf:
# Description

We want consumers to have named modules but nobody got time to do that manually.

# How to upgrade:

## Step 1

Update the pubspec.yaml in the consumer's repo to add a dependency_override:

```
  w_module:
    git:
      url: git@github.com:Workiva/w_module.git
      ref: module_name_upgrader
```

and then `pub get`.

## Step 2

In the consumer repo, run:

```
pub run w_module:name_upgrade
```

## Step 3

Commit the changes to the Dart files.

# Testing / QA

- [ ] Run it in this repo. I removed the names from all examples, so when you run it, you should see all the names added to the modules. 

- [ ] Run it in other repos with lots of modules

Note that this branch doesn't need to be merged. Just point to this branch, do the thing, and point back to whatever release the consumer was using before.


aviary3-wk:
## Security Insights

## Action Items
---

bin/name_upgrade.dart
```diff
@@ -0,0 +1,47 @@
...
+    showHelpAndExit();
+  }
+
+  if (argResults.wasParsed(helpArg) && argResults[helpArg]) {
+    showHelpAndExit();
+  }
+
+  final List<String> targets = ['lib', 'example', 'examples', 'app', 'dev-app'];
+
+  utils.moveTargetsIntoLib(targets);
```
toddbeckman-wf: Not my proudest code but this solves the problem with examples not appearing in the `AnalysisContext` and we don't have to keep running the executable over and over in subdirectories.

bin/name_upgrade.dart
```diff
@@ -0,0 +1,47 @@
...
+    showHelpAndExit();
+  }
+
+  if (argResults.wasParsed(helpArg) && argResults[helpArg]) {
+    showHelpAndExit();
+  }
+
+  final List<String> targets = ['example', 'examples', 'app', 'dev-app'];
+
+  utils.moveTargetsIntoLib(targets);
```
toddbeckman-wf: Not my proudest code but this solves the problem with examples not appearing in the AnalysisContext and we don't have to keep running the executable over and over in subdirectories.

lib/src/serializable.dart
```diff
@@ -1,239 +0,0 @@
-// Copyright 2017 Workiva Inc.
```
toddbeckman-wf: Temporarily removed because it was messing with AnalysisContext.

lib/serializable_module.dart
```diff
@@ -1,23 +0,0 @@
-// Copyright 2017 Workiva Inc.
```
toddbeckman-wf: Temporarily removed because it was messing with AnalysisContext.

example/panel/modules/data_load_async_module.dart
```diff
@@ -22,9 +22,6 @@ import 'package:w_flux/w_flux.dart';
 import 'package:w_module/w_module.dart';
 
 class DataLoadAsyncModule extends Module {
-  @override
-  final String name = 'DataLoadAsyncModule';
```
toddbeckman-wf: These and others removed in order to give an easy test case.

bin/name_upgrade.dart
dustinlessard-wf: Suggestion : I've seen this done in a few CI scripts here and there.  One common thread among all the occurrences I've seen was me asking myself 'why?'  Maybe a comment in this explaining why this is being done would be good?

bin/src/utils.dart
```diff
@@ -0,0 +1,265 @@
...
+  if (i == lines.length) {
+    print(outputLines.toString());
+    print('ran out of lines');
+    exit(1);
+  }
+
+  bool addClosingBracket = false;
+
+  // Find the line with the first {
+  for (; i < lines.length; i++) {
```
dustinlessard-wf: til

bin/src/utils.dart
```diff
@@ -0,0 +1,265 @@
...
+  final entryPoints = getPackageEntryPoints();
+  final sources = parseSources(context, entryPoints);
+  verifyAnalysis(context, sources);
+
+  return parseLibraries(context, sources)
+      .expand(getSubclassesOfLifecycleModule)
+      .where((element) => element.getGetter('name') == null);
+}
+
+void writeGettersToFile(ClassElement e) {
```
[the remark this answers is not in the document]
greglittlefield-wf: I think you could use info from the AST to reduce complexity of the insertion code:
```dart
    final source = f.readAsStringSync();
    // The offset of the end of the class's opening '{' bracket token
    final insertionOffset = (e.computeNode() as ClassDeclaration).leftBracket.end;

    outputLines
      ..write(source.substring(0, insertionOffset))
      ..writeln()
      ..writeln('  @override')
      ..writeln('  String get name => \'${e.name}\';')
      ..writeln()
      ..write(source.substring(insertionOffset));

    f.writeAsStringSync(outputLines.toString());
```

bin/src/utils.dart
```diff
@@ -0,0 +1,265 @@
...
+
+Iterable<ClassElement> getClassesThatExtendFromModule(
+    AnalysisContext context, Directory sdkDir) {
+  final entryPoints = getPackageEntryPoints();
+  final sources = parseSources(context, entryPoints);
+  verifyAnalysis(context, sources);
+
+  return parseLibraries(context, sources)
+      .expand(getSubclassesOfLifecycleModule)
+      .where((element) => element.getGetter('name') == null);
```
greglittlefield-wf: Would `getField` work here since it takes synthetic fields into account?

If so, it might be advantageous to `getGetter` since it also handles overrides of the form`final name = 'FooModule';`, but I'm not 100% on that.

greglittlefield-wf: Also #nit based on the method name, this `where` seems to belong with the consumption of the return value and not inside the method itself.

bin/src/utils.dart
```diff
@@ -0,0 +1,238 @@
...
+      results[c.source] = [c];
+    }
+  });
+
+  return results;
+}
+
+void writeGettersForSource(Source source, List<ClassElement> classes) {
+  // Sort descending to modify the bottom classes before the top
+  classes.sort((a, b) => b.computeNode().end - a.computeNode().end);
```
toddbeckman-wf: This is pretty naive but I find it works well avoiding "concurrent modification" of things like files.

greglittlefield-wf: I think it's a great solution!! 🌟

bin/src/utils.dart
```diff
@@ -0,0 +1,239 @@
...
+
+  final source = f.readAsStringSync();
+  // The offset of the end of the class's opening '{' bracket token
+  // ignore: avoid_as
+  final insertionOffset = (e.computeNode() as ClassDeclaration).leftBracket.end;
+
+  outputLines
+    ..writeln(source.substring(0, insertionOffset))
+    ..writeln('  @override')
+    ..writeln('  String get name => \'${e.name}\';')
```
greglittlefield-wf: Edge-case I just thought of: if a class name contains dollar signs, we'd want to escape them here.

We probably won't ever hit this, so we probably won't need to address it, but just in case you were feeling ambitious...

bin/name_upgrade.dart
```diff
@@ -36,15 +36,10 @@ Future main(List<String> args) async {
...
+  moveTargetsIntoLib(targets);
 
-  utils
-      .groupClassesBySource(utils.moduleClassesWithoutNames(context, sdkDir))
-      .forEach(utils.writeGettersForSource);
+  final classes = getModulesWithoutNamesBySource();
+  classes.forEach(writeGettersForSource);
 
-  utils.moveTargetsOutOfLib(targets);
+  moveTargetsOutOfLib(targets);
```
greglittlefield-wf: #nit Just thought... Could we put this in a `finally` block so just in case something goes wrong, it doesn't leave the repo in a bad state?

pubspec.yaml
```diff
@@ -11,6 +11,8 @@ authors:
 homepage: https://github.com/Workiva/w_module
 
 dependencies:
+  analyzer: ^0.29.11
```
greglittlefield-wf: Dependency issues here as well:
>w_module from path depends on analyzer ^0.29.11 and dart_medic >=1.3.6 depends on analyzer ^0.30.0

So, this should be `^0.30.0`, then?

bin/name_upgrade.dart
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  final List<String> targets = ['example', 'examples', 'app', 'dev-app'];
+
+  // The AnalyzerContext we use only contains the content in lib.
+  // Move all the code we care about into lib temporarily to upgrade any
+  // modules that might live outside of there normally.
+  moveTargetsIntoLib(targets);
+
+  final classes = getModulesWithoutNamesBySource();
```
greglittlefield-wf: #nit Could we add some messaging here just so people running this script know it will take a while?
```diff
+print('Analyzing sources to find modules to upgrade.');
+print('(This can take several minutes for large repos)');
 final classes = getModulesWithoutNamesBySource();
+print('Upgrading modules...');
 classes.forEach(writeGettersForSource);
```

pubspec.yaml
toddbeckman-wf:
```
    conflict: dart_style >=0.2.15 <1.0.2-dev depends on analyzer ^0.29.5
    ! not analyzer ^0.29.5 is satisfied by analyzer ^0.30.0
    ! which is caused by "w_module depends on analyzer ^0.30.0"
    ! thus: dart_style >=0.2.15 <1.0.2-dev is forbidden
    ! dart_style >=0.2.15 <1.0.2-dev is satisfied by dart_style ^0.2.16
    ! which is caused by "w_module depends on dart_style ^0.2.16"
    ! thus: version solving failed
Version solving took 0:00:04.756626 seconds.
Tried 1 solutions.
Because dart_style >=0.2.15 <1.0.2-dev depends on analyzer ^0.29.5 and w_module depends on
  analyzer ^0.30.0, dart_style >=0.2.15 <1.0.2-dev is forbidden.
So, because w_module depends on dart_style ^0.2.16, version solving failed.
```

toddbeckman-wf:
That issue seems to be because `.pub` is getting copied. I'll make it not do `app`/`dev-app` to avoid nested repo issue, but it will continue happening in `wdesk_sdk` until you `pubclean` in `example` due to that being a nested project.

bin/name_upgrade.dart
toddbeckman-wf: I moved it to a finally but it didn't fix the bad state 🤷

patkujawa-wf:
Are there plans to run this on our repos with repo Commander or Bender?

PR #137, opened 2018-10-24
AF-2971 Dart 2 compatibility

robbecker-wf:
# Overview
Updates are needed to use this library under Dart 2

# Changes
 - updated analysis options and renamed to the proper name
 - updated dependencies and dart sdk ranges in pubspec.yaml
 - unit tests fixed for dart 2.
 - fixed some type errors in the examples
 - CI now enforces dartfmt under dart 2

# Testing
 - [ ] CI passes under dart 1 and dart 2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM google/dart:1.24.3 as build
-
-COPY pubspec.* /build/
+WORKDIR /build/
+ADD pubspec.* /build/
 RUN pub get 
 ARG BUILD_ARTIFACTS_BUILD=/build/pubspec.lock
 FROM scratch
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,43 +1,6 @@
 FROM google/dart:1.24.3 as build
 
-RUN apt-get update -qq
-RUN apt-get update && apt-get install -y \
-	build-essential \
-	wget \
-	&& rm -rf /var/lib/apt/lists/*
-
-# install chrome
-RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - && \
-	echo 'deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main' | tee /etc/apt/sources.list.d/google-chrome.list && \
-	apt-get -qq update && apt-get install -y google-chrome-stable && \
-	mv /usr/bin/google-chrome-stable /usr/bin/google-chrome && \
-	sed -i --follow-symlinks -e 's/\"\$HERE\/chrome\"/\"\$HERE\/chrome\" --no-sandbox/g' /usr/bin/google-chrome && \
-	google-chrome --version
-
-
-WORKDIR /build
-
-# setup ssh
-ARG GIT_SSH_KEY
-ARG KNOWN_HOSTS_CONTENT
-RUN mkdir /root/.ssh/ && \
-  echo "$KNOWN_HOSTS_CONTENT" > "/root/.ssh/known_hosts" && \
-  chmod 700 /root/.ssh/ && \
-  umask 0077 && echo "$GIT_SSH_KEY" >/root/.ssh/id_rsa && \
-  eval "$(ssh-agent -s)" && ssh-add /root/.ssh/id_rsa
-
-# grab source
-COPY . /build/
-
-# deps / test / sanity check / build
-ENV DART_FLAGS="--checked"
-ARG BUILD_ID
-ARG GIT_COMMIT
-ARG GIT_TAG
-ARG GIT_BRANCH
-ARG GIT_MERGE_HEAD
-ARG GIT_MERGE_BRANCH
-RUN pub get --packages-dir && \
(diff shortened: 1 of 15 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

example/panel/modules/sample_tracer.dart
```diff
@@ -32,8 +32,7 @@ class SampleSpan implements Span {
     this.references,
     DateTime startTime,
     Map<String, dynamic> tags,
-  })
-      : this.startTime = startTime ?? new DateTime.now(),
+  })  : this.startTime = startTime ?? new DateTime.now(),
```
robbecker-wf: dart 2 formatter change

lib/src/lifecycle_module.dart
```diff
@@ -594,7 +594,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
           ]);
     }
 
-    Future pendingTransition;
+    Future<Null> pendingTransition;
```
robbecker-wf: changed to match the type signature of _suspend

lib/src/lifecycle_module.dart
```diff
@@ -661,7 +661,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
           allowedStates: [LifecycleState.suspended, LifecycleState.suspending]);
     }
 
-    Future pendingTransition;
+    Future<Null> pendingTransition;
```
robbecker-wf: changed to match the type signature of _resume

test/lifecycle_module_test.dart
```diff
@@ -280,6 +282,8 @@ void expectInLifecycleState(LifecycleModule module, LifecycleState state) {
 }
 
 Future<Null> gotoState(LifecycleModule module, LifecycleState state) async {
+  // wait for next event loop. fixes sync-async in Dart 2
+  await new Future.value(null);
```
robbecker-wf: @Workiva/app-frameworks want to take a look at these closely to determine if we're going to have a sync-async problem with lifecycle module or if it's just the tests and mocks?

Dockerfile
```diff
@@ -1,43 +1,6 @@
 FROM google/dart:1.24.3 as build
-
```
robbecker-wf: Just let Travis CI run all the tests.

test/lifecycle_module_test.dart
robbecker-wf: well, goto state is only used internally in tests, so I'm not too concerned about this one.

test/lifecycle_module_test.dart
```diff
@@ -149,6 +149,7 @@ class TestLifecycleModule extends LifecycleModule {
   @override
   @protected
   Future<Null> onWillUnloadChildModule(LifecycleModule module) async {
+    await new Future.value(null);
```
robbecker-wf: This spot is a little more concerning because this may mean that it's possible for other consumers that are overriding one of these lifecycle methods may get a change in how lifecycle events are ordered due to sync-async. I think we can go ahead with this PR as is, but be aware that there may be a recommended migration step of putting a line like this in each overriden module lifecycle method.

lib/src/lifecycle_module.dart
robbecker-wf: this is internal and not a public api change

lib/src/lifecycle_module.dart
robbecker-wf: also internal only. no public api change.

test/lifecycle_module_test.dart
robbecker-wf: Without this await here .. the order that events are added to the tracking list of events is different than what the tests expect the order to be. This looks like:

```
00:05 +216 -1: [VM] test/lifecycle_module_test.dart: LifecycleModule with globalTracer with children unload should wait for in-progress child module loads  [E]
  Expected: [
              'onWillLoadChildModule',
              'willLoadChildModule',
              'onShouldUnload',
              'willUnload',
              'onDidLoadChildModule',
              'didLoadChildModule',
              'onWillUnloadChildModule',
              'willUnloadChildModule',
              'onDidUnloadChildModule',
              'didUnloadChildModule',
              'onUnload',
              'didUnload',
              'onDispose'
            ]
    Actual: [
              'onWillLoadChildModule',
              'willLoadChildModule',
              'onShouldUnload',
              'willUnload',
              'onDidLoadChildModule',
              'onWillUnloadChildModule',
              'didLoadChildModule',
              'willUnloadChildModule',
              'onDidUnloadChildModule',
              'didUnloadChildModule',
              'onUnload',
              'didUnload',
              'onDispose'
            ]
     Which: was 'onWillUnloadChildModule' instead of 'didLoadChildModule' at location [5]

  package:test                             expect
  test/lifecycle_module_test.dart 2236:11  runTests.<fn>.<fn>.<fn>.<fn>
```

evanweible-wf:
QA +1
- [x] CI passes on Dart 1 and Dart 2

@Workiva/release-management-p

PR #139, opened 2018-12-11
Clear child modules on dispose.

joshprzybyszewski-wf:
## Problem
We're seeing modules get leaking text_doc_client.

## Solution
We think that clearing the list of modules is a fail-safe solution to ensure that we don't retain references to children modules when we've been disposed.

@jacobjohnson-wf @evanweible-wf @bencampbell-wf

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -514,6 +514,14 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
           onError: (error, stackTrace) =>
               _didUnloadChildModuleController.addError);
 
+      // The child module may not reach an unloaded state successfully, but
+      // should always eventually be disposed. For this reason, we listen for
+      // its disposal before removing it from the list of child modules.
+      // ignore: unawaited_futures
+      childModule.didDispose.then((_) {
+        _childModules.remove(childModule);
+      });
+
       try {
         manageDisposable(childModule);
         _childModules.add(childModule);
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -483,7 +483,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
   /// Attempting to load a child module after a module has been unloaded will
   /// throw a [StateError].
   @protected
-  Future<Null> loadChildModule(LifecycleModule childModule) {//
+  Future<Null> loadChildModule(LifecycleModule childModule) {
     if (isOrWillBeDisposed) {
       return _buildDisposedOrDisposingResponse(methodName: 'loadChildModule');
     }
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -483,7 +483,7 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
   /// Attempting to load a child module after a module has been unloaded will
   /// throw a [StateError].
   @protected
-  Future<Null> loadChildModule(LifecycleModule childModule) {
+  Future<Null> loadChildModule(LifecycleModule childModule) {//
     if (isOrWillBeDisposed) {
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Security Insights

## Action Items
---
_Questions or Comments? Reach out on Slack: #support-infosec._

lib/src/lifecycle_module.dart
```diff
@@ -889,6 +889,17 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
...
 
+  @mustCallSuper
+  @override
+  @protected
+  Future<Null> onDispose() async {
+    if (isInstantiated || isUnloaded) {
+      return;
+    }
+
+    _childModules.clear();
```
evanweible-wf: We should add a test for this. I think the easiest way to do that would be to insert a `tearDown` step right here:
https://github.com/Workiva/w_module/blob/master/test/lifecycle_module_test.dart#L1502

```dart
tearDown(() {
  if (withChild) {
    expect(module.childModules, isEmpty);
  }
});
```

joshprzybyszewski-wf:
Hold on this; this might be a good change, but we might have isolated our problem to a different location.

bencampbell-wf:
We remove child modules on disposal of the child module. I would suggest that we remove child modules on the unload of the child module as well.

https://github.com/Workiva/w_module/blob/f8b3602c3e355d6a240d183116c55e1f5408e99b/lib/src/lifecycle_module.dart#L518

lib/src/lifecycle_module.dart
```diff
@@ -889,6 +889,17 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
     }
   }
 
+  @mustCallSuper
```
greglittlefield-wf: So that consumers who override `onDispose` and don't already call super, should we put this cleanup logic somewhere else?

Perhaps in the constructor:
```dart
getManagedDisposer(() async {
  if (isInstantiated || isUnloaded) {
    return;
  }
  _childModules.clear();
});

File: lib/src/lifecycle_module.dart (hunk 1 above)
Comment (todbachman-wf) [reply lib/src/lifecycle_module.dart]: I would have expected to see this right after line 888 where we are disposing of the child modules. It seems like you guys have found a bug that is preventing us from hitting line 888 where we dispose of the child modules--is anyone trying to figure out why that's not getting hit?

File: lib/src/lifecycle_module.dart [hunk 3]
```diff
@@ -510,15 +514,8 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
           onError: (error, stackTrace) =>
               _didUnloadChildModuleController.addError);
 
-      // The child module may not reach an unloaded state successfully, but
-      // should always eventually be disposed. For this reason, we listen for
-      // its disposal before removing it from the list of child modules.
-      // ignore: unawaited_futures
-      childModule.didDispose.then((_) {
-        _childModules.remove(childModule);
-      });
```
greglittlefield-wf: I think we need to keep this in place so that unloaded child modules aren't retained in `_childModules` until the parent is unloaded (not sure why we don't have a test for that 😞). 

If we restore this, we should be able to remove the `_childModules.clear()` disposer, but I wouldn't be opposed to keeping it around to be safe.

evanweible-wf:
@joshprzybyszewski-wf @jacobjohnson-wf would you guys be able to use this branch to verify that it fixes the memory leaks you were seeing? That would serve as a good QA for this PR

evanweible-wf:
^ nevermind, looking at the DT PR again, it looks like the leak would be addressed without this PR anyway, so it's not a great candidate for QA here.

greglittlefield-wf:
- Changes look good
- All cases are covered by unit tests
- CI passes, and tests verify that children are neither retained via `_childModules` and arre disposed properly

QA +1

PR #143, opened 2019-04-23
CPLAT-5450: Updates to better support the first render use case

paulgerlich-wf:
## Motivation
First useful should be a publicly subscript-able event that parents can listen for.

#### Release Notes
Parent containers can now listen to child modules for their first useful events in better tracking component initial renders. ?? idk

## Review
Please review: @greglittlefield-wf @evanweible-wf 

### QA Checklist
- [] Tests were updated and provide good coverage of the changeset and other affected code
**Do you want me to do anything here?**
- [ ] Manual testing was performed if needed

## Merge Checklist
While we perform many automated checks before auto-merging, some manual checks are needed:
- [ ] A Client Platform member has reviewed these changes
- [ ] There are no unaddressed comments _- this check can be automated if reviewers use the "Request Changes" feature_
- [ ] _For release PRs -_ Version metadata in Rosie comment is correct

[contributing-review-types]: https://github.com/Workiva/w_module/blob/master/CONTRIBUTING.md#review-types
[contributing-manual-testing]: https://github.com/Workiva/w_module/blob/master/CONTRIBUTING.md#manual-testing-criteria

### DIFF
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -212,12 +212,27 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
   void specifyFirstUsefulState({
     Map<String, dynamic> tags: const {},
     List<Reference> references: const [],
-    DateTime trueStartTime,
   }) {
     specifyStartupTiming(StartupTimingType.firstUseful,
-        tags: tags, references: references, trueStartTime: trueStartTime);
+        tags: tags, references: references);
+  }
 
-    _isFirstUsefulController.add(this);
+  /// Creates a span with `globalTracer` starting now
+  /// (or at an optional [startTime]).
+  ///
+  /// This span is intended to represent the time it takes for the module to
+  /// finish asynchronously loading any necessary data and entering a state which
+  /// is ready for user interaction.
+  ///
+  /// Any [tags] or [references] specified will be added to this span.
+  @protected
+  Span startFirstUsefulStateSpan({
+    Map<String, dynamic> tags: const {},
+    List<Reference> references: const [],
+    DateTime startTime,
+  }) {
+    return startStartupTimingSpan(StartupTimingType.firstUseful,
+        tags: tags, references: references, startTime: startTime);
   }
 
   /// Creates a span with `globalTracer` from the start of [load] until now.
@@ -230,15 +245,33 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
     StartupTimingType specifier, {
     Map<String, dynamic> tags: const {},
     List<Reference> references: const [],
-    DateTime trueStartTime,
   }) {
-    // Start time for load was overridden in the case of deferred module rendering
-    DateTime startTime = trueStartTime ?? _startLoadTime;
-
     // Load didn't start
-    if (_loadContext == null || startTime == null) {
+    if (_startLoadTime == null) {
       throw new StateError(
           'Calling `specifyStartupTiming` before calling `load()`');
+    }
+    startStartupTimingSpan(specifier,
+            tags: tags, references: references, startTime: _startLoadTime)
+        ?.finish();
+  }
+
+  /// Creates a span with `globalTracer` starting now
+  /// (or at an optional [startTime].
+  ///
+  /// The [specifier] indicates the purpose of this span.
+  ///
+  /// Any [tags] or [references] specified will be added to this span.
+  Span startStartupTimingSpan(
+    StartupTimingType specifier, {
+    Map<String, dynamic> tags: const {},
+    List<Reference> references: const [],
+    DateTime startTime,
+  }) {
+    // Load didn't start
+    if (_loadContext == null) {
+      throw new StateError(
+          'Calling `startStartupTimingSpan` before calling `load()`');
     }
 
     final tracer = globalTracer();
@@ -246,19 +279,24 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       return null;
     }
 
-    _logger.finest(
-        'specifyStartupTiming for $name. It took ${new DateTime.now().difference(startTime).inSeconds} seconds for this module to be interactable.');
+    final span = tracer.startSpan(
+      '$name.${specifier.operationName}',
+      references: [tracer.followsFrom(_loadContext)]..addAll(references),
+      startTime: startTime,
+      tags: _defaultTags..addAll(tags),
+    );
 
-    tracer
-        .startSpan(
-          '$name.${specifier.operationName}',
-          references: [tracer.followsFrom(_loadContext)]..addAll(references),
-          startTime: startTime,
-          tags: _defaultTags..addAll(tags),
-        )
-        .finish();
+    switch (specifier) {
+      case StartupTimingType.firstUseful:
+        span.whenFinished.then((_) {
+          _isFirstUsefulController.add(this);
+          _logger.finest(
+              'firstUseful for $name. It took ${span.duration.inSeconds} seconds for this module to be interactable.');
+        });
+        break;
+    }
 
-    _startLoadTime = null;
+    return span;
   }
 
   /// Name of the module for identification in exceptions and debug messages.
--- a/lib/src/lifecycle_module.dart
+++ b/lib/src/lifecycle_module.dart
@@ -72,36 +72,36 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
 
   // Lifecycle event StreamControllers
   StreamController<LifecycleModule> _willLoadChildModuleController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
   StreamController<LifecycleModule> _didLoadChildModuleController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
 
   StreamController<LifecycleModule> _willLoadController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
   StreamController<LifecycleModule> _didLoadController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
   StreamController<LifecycleModule> _isFirstUsefulController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
 
   StreamController<LifecycleModule> _willSuspendController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
   StreamController<LifecycleModule> _didSuspendController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
 
   StreamController<LifecycleModule> _willResumeController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
   StreamController<LifecycleModule> _didResumeController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
 
   StreamController<LifecycleModule> _willUnloadChildModuleController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
   StreamController<LifecycleModule> _didUnloadChildModuleController =
-  new StreamController<LifecycleModule>.broadcast();
+      new StreamController<LifecycleModule>.broadcast();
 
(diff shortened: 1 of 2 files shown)


paulgerlich-wf:
Just a sec, test updates incoming.

aviary-wf:
## Security Insights

## Action Items
---
_Questions or Comments? Reach out on Slack: #support-infosec._

lib/src/lifecycle_module.dart
```diff
@@ -159,6 +163,9 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
   @protected
   Span get activeSpan => _activeSpan;
 
+  /// If this module has had an overridden name and therefor is being traced
+  bool get isTraced => name != _defaultName;
```
paulgerlich-wf: There was no way to know that a module is actually being traced externally, unless we tried to guess at this default name. We figured a private string is liable to change, though, so it's better to expose this property to consumers.

lib/src/lifecycle_module.dart
```diff
@@ -205,12 +212,13 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
   void specifyFirstUsefulState({
     Map<String, dynamic> tags: const {},
     List<Reference> references: const [],
-  }) =>
-      specifyStartupTiming(
-        StartupTimingType.firstUseful,
-        tags: tags,
-        references: references,
-      );
+    DateTime trueStartTime,
```
paulgerlich-wf: Definitely open to renaming this. Essentially this allows the module to specify that the "true start time" (i.e, the actual wait a consumer perceives) isn't necessarily the moment that the module's `load` function is called.

lib/src/lifecycle_module.dart
```diff
@@ -234,11 +246,14 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
       return null;
     }
 
+    _logger.finest(
```
paulgerlich-wf: Open to killing this log, but it's pretty nice to have in your console.

lib/src/lifecycle_module.dart
greglittlefield-wf: What's the use-case for telling whether a module is traced?

Also, this will only ever be `true` in minified dart2js, and not other runtimes.

lib/src/lifecycle_module.dart
```diff
@@ -272,6 +287,12 @@ abstract class LifecycleModule extends SimpleModule with Disposable {
   /// [onLoad] call will be emitted.
   Stream<LifecycleModule> get didLoad => _didLoadController.stream;
 
+  /// The [LifecycleModule] has been declared firstUseful
+  ///
+  /// Used for the parent module to know that its child is useful, as oftentimes
+  /// a parent can't be called "Useful" until all of its children are
+  Stream<LifecycleModule> get isFirstUseful => _isFirstUsefulController.stream;
```
greglittlefield-wf: Do you have an example of how this will be used in practice?

lib/src/lifecycle_module.dart
paulgerlich-wf: Name won't equal default name if we override the name property with a hard coded string, right? My understanding was that overriding the default name with a hard coded string would allow the module to be traced with a meaningful span prefix.

Here is our use case, where a parent listens to both module isFirstUseful and manual spans to determine that it is itself firstUseful:
https://github.com/Workiva/graph_app/pull/7145/files#diff-67c7c5d0ea4739dbee0a0091e12a0dbdR83

lib/src/lifecycle_module.dart
evanweible-wf: Could you provide an example where this would result in a meaningful difference? I'm concerned with adding an option for overriding the start time because we want to provide as much of this profiling information out of the box as possible. If we don't have a well-defined, consistent start time, the metrics between experiences may vary considerably based on how `trueStartTime` is used.

lib/src/lifecycle_module.dart
evanweible-wf: Just curious, what's the use case for this?

lib/src/lifecycle_module.dart
evanweible-wf: Should this actually be a stream or just a future?

lib/src/lifecycle_module.dart
greglittlefield-wf: Only seconds granularity? That's rather pessimistic 😉 

Haha but for real, I think more precision on this might be nice ("took 1.2s"/"took 1200ms"/etc.)

lib/src/lifecycle_module.dart
paulgerlich-wf: Here is our use case: https://github.com/Workiva/graph_app/pull/7145/files#diff-67c7c5d0ea4739dbee0a0091e12a0dbdR83

The idea is that in practice we don't have a perfect modular hierarchy. Graph_app parent modules might have children that are modules and children that are strictly dumb components, but we don't want to specify the overarching parent module as useful until both the child modules and dumb components are "first useful", so we track both manual spans and child.isFirstUseful

paulgerlich-wf: That's a good point. A future is a better solution

lib/src/lifecycle_module.dart
paulgerlich-wf: Totally agree, I just started off with something useful for debugging but I can clean this up.

lib/src/lifecycle_module.dart
paulgerlich-wf: Yeah so in a sox test form we load up the first Phase module and overview module. The user is on the overview for an indeterminate amount of time, but the phase module's onLoad has already been called. It could be 5 minutes before they actually open the phase, but the only time we care about is the time from them clicking that tab -> first render of the components / modules in that tab.

lib/src/lifecycle_module.dart
paulgerlich-wf: It's for tracking child module spans to determine the parent is first useful. If the module didn't have the `name` property overriden, we'd have no way to know that it's impossible for that module to trigger the `isFirstUseful` so our parent's `isFirstUseful` would also never trigger.

lib/src/lifecycle_module.dart
georgelesica-wf: Do we support unloading and reloading a module? My recollection is that we punted on that a long time ago, although now with Disposable in the mix we probably really don't support that. So realistically this should probably be a future.

lib/src/lifecycle_module.dart
georgelesica-wf: Wait, huh? A module name should only ever be informative for logging purposes, we shouldn't be relying on that for behavior in code. What if it changes?

paulgerlich-wf: This is the origination: https://github.com/Workiva/w_module/blob/master/lib/src/lifecycle_module.dart#L175

It's a necessary evil since we obfuscate the JS. Without it, we'd have no meaningful naming for the lifecycle event spans.

lib/src/lifecycle_module.dart
paulgerlich-wf: For sure, I was considering taking a different approach at first but I realized our use case of the LifecycleModule is probably both wrong and unusual so the single-load / future approach is definitely the way to go 😄

lib/src/lifecycle_module.dart
paulgerlich-wf: To be clear: We don't care what the name is necessarily, we just care if it is set because that means that it's being traced and we're able to feed it's "firstUseful" into the parents "firstUseful"

greglittlefield-wf: Ah, my bad, I was mistaken; the default name of is `LifecycleModule(${this.runtimeType})`, so you're correct in that any named module will be traced.

@georgelesica-wf we disable tracing when a custom name has not been set, because otherwise the minified name would be used in dart2js and you'd get traces with names like `LifecycleModule(jfkdlas)`.

lib/src/lifecycle_module.dart
greglittlefield-wf: Based on that use-case, I would ask:
1. Why is the module being loaded if it's not being used
2. This timing is defined as the span between load and interactive; as opposed to changing this definition, would a custom tracing span to capture the time between "phase opened" and "interactive" be a better fit?

paulgerlich-wf: 1. It's preloading part of the data so that it's partially available when the user switches tabs, instead of cold starting on every tab switch (this significantly improved load times for us some time ago).
 
2. You're right. Perhaps we could modify that definition because it assumes a relatively rigid use of the module? Generally what we're trying to track is perceived load time by the user, not necessarily whatever the component lifecycle ends up being. The onLoad span tracks the lifecylce, the firstUseful should be more ad-hoc and defined by the module implementation IMO.

I think we could probably adjust the definition to:
```
  /// Creates a span with `globalTracer` to track the perceived load time of this module.
  ///
  /// This span is intended to represent the time it takes for the module to
  /// finish asynchronously loading any necessary data and entering a state which
  /// is ready for user interaction.
```

Thoughts?

lib/src/lifecycle_module.dart
greglittlefield-wf: This might actually be better as a Stream since subscriptions can be cleanly cancelled, unlike with futures.

We'll likely want this, as we can't assume that:
- all traced modules will emit this event
- traced modules will emit this event before the listener is disposed

Also, if it were a stream, we could just make it be an empty stream when the module is not traced and thus will never be called, causing the subscription to immediately be cancelled.

```dart
listenToStream(childModule.isFirstUseful.take(1), (module) {
  // ...
});
```
vs
```dart
childModule.isFirstUseful.then((module) {
  if (iAmDisposedOrThisThenBlockIsOtherwiseInvalid) return;
  // ...
});
```

lib/src/lifecycle_module.dart
paulgerlich-wf: After re-reading your second question, I think you're right. We could just define a manual span that captures what we want to capture. That would remove the need for us to overwrite this start time here. 

We would still want the isFirstUseful future and "isTraced" bool property though. Does that sound good?

greglittlefield-wf: I actually think your definition makes a lot more sense; we defined these without having very clear use-cases, so it makes sense to change them to match how users want to use them.

I'd be open to that definition changing, but I think it'd be nice for the new "start" point to be defined via an API closer to the creation of a span, as opposed to passing in a time.

For example, instead of :
```dart
var startTime = new DateTime.now();
// ...
specifyFirstUsefulState(startTime: startTime);
```
```dart
final firstUsefulSpan = startFirstUsefulStateSpan();
//
firstUsefulSpan.finish();
```

This has the advantage of recording these spans as marks/measures, and keeps managing of the start/end points of the span more consistent with other tracing APIs.

---

> We would still want the isFirstUseful future and "isTraced" bool property though. Does that sound good?

👍

paulgerlich-wf: That would be awesome, I would prefer that as well because the dart tracing console doesn't show spans with start times that were injected. So they just end up being 1ms spans (at least visually, but in on our tracing dashboard).

The only issue is that some of the module's children are preloaded during the onLoad. We want both those children that were pre-loaded (so loaded before the startFirstUsefulStateSpan()) and the ones called during the first render cycle to share the same parent context.

The current method allows both the pre-loaded and loaded-on-demand spans to share a parent context. Does that make sense? Any thoughts for a way to fix that issue?

lib/src/lifecycle_module.dart
paulgerlich-wf: Aha I didn't know about .take(1), that's perfect. Thanks Greg.

lib/src/lifecycle_module.dart
georgelesica-wf: Oh OK, gotcha, I wasn't aware there was anything related to tracing built-in here.

lib/src/lifecycle_module.dart
georgelesica-wf: Can we just expose the `take(1)` stream so consumers don't have to know to do that?

greglittlefield-wf: Part of me likes the convenience of building that in, but another part of me thinks that it obscures the actual behavior and discourages defensive/robust stream listening practices.

georgelesica-wf: Obscures the actual behavior? It's a stream that will never issue more than one event. Isn't that the actual behavior?

lib/src/lifecycle_module.dart
greglittlefield-wf: As per offline discussion, this has been addressed in https://github.com/Workiva/w_module/pull/143/commits/9520d8704d36f43259b5c89efc29bc3171328443

PR #267, opened 2025-07-18
Rollout React 18 to tests and example apps

sourcegraph-wk:
This PR updates all references to React JS files from the React 17 to [React 18 versions](https://github.com/Workiva/react-dart?tab=readme-ov-file#react-18) which will cause most tests and example apps to run on React 18.


todbachman-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

